PR #141, opened 2023-05-26
Fix possible Actionjournal errors on znr

forsthug:

{Summary of the changes}

## Description

{Detail}

Fixes #{issue number}

### DIFF
--- a/queue/src/fiskaltrust.Middleware.Localization.QueueDE/Repositories/ReadOnlyReceiptReferenceRepository.cs
+++ b/queue/src/fiskaltrust.Middleware.Localization.QueueDE/Repositories/ReadOnlyReceiptReferenceRepository.cs
@@ -125,7 +125,7 @@ namespace fiskaltrust.Middleware.Localization.QueueDE.Repositories
 
         private async Task<(long, DateTime?)> GetLastZNumberForQueueItem(ftQueueItem queueItem)
         {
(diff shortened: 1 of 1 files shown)


queue/src/fiskaltrust.Middleware.Localization.QueueDE/Repositories/ReadOnlyReceiptReferenceRepository.cs
```diff
@@ -124,19 +125,23 @@ private async Task<bool> AddReference(HashSet<ReceiptReferencesGroupedData> rece
 
         private async Task<(long, DateTime?)> GetLastZNumberForQueueItem(ftQueueItem queueItem)
         {
-            var actionJournals = (await _actionJournalRepository.GetAsync()).Where(x => x.Type == "4445000000000007").OrderBy(x => x.TimeStamp);
+            var regexDaily = new Regex("4445[0-9]{11}7");
```
volllly, line 128:
```suggestion
            var regexDaily = new Regex("4445[0-9]{8}0007");
```
I think technically the last 4 numbers are part of the receipt case. And we also have `0x4445000000000017` (init-scu-switch) right now which would be caught wrongly with the current regex

volllly (changes requested):

volllly (approved):

PR #145, opened 2023-05-31
Fix nuspec file

volllly:


TSchmiedlechner:
@volllly I think we won't need this anymore 😁

PR #150, opened 2023-06-14
57830 - SCU-Switch force switch flag 

forsthug:

{Summary of the changes}

## Description

{Detail}

Fixes #{issue number}


queue/src/fiskaltrust.Middleware.Localization.QueueDE/Extensions/ReceiptRequestExtensions.cs
```diff
@@ -74,6 +74,10 @@ public static bool IsRemoveOpenTransactionsWhichAreNotOnTse(this ReceiptRequest
         {
             return ((receiptRequest.ftReceiptCase & 0x0000_0000_2000_0000) > 0x0000);
         }
+        public static bool IsInitiateScuSwitchReceiptForce(this ReceiptRequest receiptRequest)
+        {
+            return ((receiptRequest.ftReceiptCase & 0x0000_0000_0010_0000) > 0x0000);
```
volllly, line 79: We have `0x0000_0000_0010_0000` also documented as "Small business, not taxable sales." but it is not actually used in the DE Queue. Maybe we should remove it from the docs or change the flag in this case or both 🤷

But we also have used the flag `0x0000_0000_0100_0000` for three different purposes so I guess its not a problem If there is no case in which they would be used at the same time 🤔

volllly (approved):

PR #157, opened 2023-06-15
overloads for the new v2 api

### LINKED ISSUE #146: CryptoVision v2 Support
Since the CryptoVision v2 has been released (https://www.cryptovision.com/de/cryptovision-tsev2-erhaelt-bsi-zertifikat/) we should validate which changes we do need to make the Middleware work with the new firmware.

This issue should summarize the necessary steps and design decisions that we are making for adding support. As a first step we need to analyze the new interface description and compare it to our current implementation to see which changes are required.

Beside that we will need to check if there are any conceptual changes to the v2 and if we are still able to use File I/O the way we are currently doing it.

### Todos

- [ ] Analyze new documentation and compare with current implementation
- [ ] Check if the current implementation will be extended or if we should create a new package
- [ ] Prepare documentation on how to switch v1 to v2 
- [ ] The new legislation requires us to collect info on the connection to the so called Sperrliste we should check how we can get this information

paulcristiann:
- [] You've included unit or integration tests for your change, where applicable.
- [X] You've included inline docs for your change, where applicable.
- [X] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

{Summary of the changes}

## Description
Added overloads for the InitializePin function of the CryptoVision SCU

{Detail}
The overloads take less parameters and should be only called with V2 hardware and firmware

Fixes #146

### DIFF
--- a/scu-de/src/fiskaltrust.Middleware.SCU.DE.CryptoVision/Interop/File/CryptoVisionFileProxy.cs
+++ b/scu-de/src/fiskaltrust.Middleware.SCU.DE.CryptoVision/Interop/File/CryptoVisionFileProxy.cs
@@ -427,6 +427,20 @@ namespace fiskaltrust.Middleware.SCU.DE.CryptoVision.Interop
             });
         }
 
+        /// <summary>
+        /// This is the overload of the InitializePinAsync command which is used ONLY with V2 hardware
+        /// </summary>
+        public async Task<SeResult> SeInitializePinsAsync(string userId, byte[] userPin)
+        {
+            return await CommandRunner.ExecuteAsync(async () =>
+            {
+                var command = AdditionalCommands.CreateInitializePinsTseCommand(userId, userPin);
+                var response = await _transportAdapter.ExecuteAsync(command);
+
+                return SeResult.ExecutionOk;
+            });
+        }
+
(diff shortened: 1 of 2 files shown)


paulcristiann:

paulcristiann:
recheck

scu-de/src/fiskaltrust.Middleware.SCU.DE.CryptoVision/Interop/File/CryptoVisionFileProxy.cs
```diff
@@ -427,6 +427,20 @@ public async Task<SeResult> SeInitializePinsAsync(byte[] adminPuk, byte[] adminP
             });
         }
 
+        /// <summary>
+        /// This is the overload of the InitializePinAsync command which is used ONLY with V2 hardware
+        /// </summary>
+        public async Task<SeResult> SeInitializePinsAsync(string userId, byte[] userPuk)
```
StefanKert, line 433: Looks like currently we are not calling this from our CryptoVisionSCU. We should add this call for V2 based TSEs

paulcristiann:
recheck

StefanKert (approved):

PR #164, opened 2023-06-22
Add_Maximum_Pool_Size_config_parameter_to_MySQL_queue

pawelvds:


queue/src/fiskaltrust.Middleware.Storage.MySQL/MySQLStorageConfiguration.cs
```diff
@@ -7,6 +7,9 @@ public class MySQLStorageConfiguration
     {
         [JsonProperty("connectionstring")]
         public string ConnectionString { get; set; }
+        
+        [JsonProperty("maxpoolsize")]
```
volllly, line 11: You can leave ths is out because `JsonConvert.DeserializeObject` is case insensitive (IDK why we've set that for the connectionstring ^^)

volllly (approved):

PR #168, opened 2023-07-03
Default queue

pawelvds:

{Summary of the changes}

## Description

{Detail}

Fixes #{issue number}


queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/PosReceiptCommand.cs
```diff
@@ -0,0 +1,243 @@
...
+                });
+                await _journalITRepository.InsertAsync(journalIT).ConfigureAwait(false);
+            }
+
+            return new RequestCommandResponse
+            {
+                ReceiptResponse = receiptResponse,
+                Signatures = receiptResponse.ftSignatures.ToList(),
+                ActionJournals = new List<ftActionJournal>()
+            };
```
volllly, line 100:
```suggestion
            var queueIt = await _countrySpecificQueueRepository.GetQueueAsync(queue.ftQueueId).ConfigureAwait(false);
            
            return new RequestCommandResponse
            {
                ReceiptResponse = CreateReceiptResponse(queue, request, queueItem, queueIt.CashBoxIdentification, _countryBaseState),
                Signatures = new(),
                ActionJournals = new()
            };
```

I think you can just return it like that and remove everything above 🤔

TSchmiedlechner:
Didn't review this yet in detail, so more of a general question: what's the "goal" of having this in place? Will we communicate this to our users, and if yes, how? Or is it meant to be for internal usage only so that we can test things quicker?

queue/src/fiskaltrust.Middleware.Queue/Bootstrapper/LocalizedQueueBootStrapperFactory.cs
```diff
@@ -27,7 +28,7 @@ public static ILocalizedQueueBootstrapper GetBootstrapperForLocalizedQueue(Guid
                 "FR" => middlewareConfiguration.PreviewFeatures.TryGetValue("queue-fr", out var val) && val ? new QueueFRBootstrapper() : throw new NotImplementedException("The French Queue is not yet implemented in this version."),
                 "IT" => new QueueITBootstrapper(),
                 "ME" => new QueueMeBootstrapper(),
-                _ => throw new ArgumentException($"Unkown country code: {countyCode}"),
+                _ => new QueueDEFAULTBootstrapper(),
```
StefanKert, line 31: IMO we should be careful here in case of future upgrades from the Queue.

e.g. we implement Croatia (countrycode HR) and someone is already using croatia with the generic queue. In this case a package upgrade that includes a working version for HR would cause a breaking change. So maybe we should only allow this "Default" implementation when explicitly defined?

TSchmiedlechner: Agree 👍🏻 and can we please limit this to sandbox only? Production Queues should definitely not be allowed to use this in any case.

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/QueueDEFAULTBootstrapper.cs
```diff
@@ -22,8 +24,18 @@ public void ConfigureServices(IServiceCollection services)
                 .AddScoped<SignatureItemFactoryDEFAULT>()
                 .AddScoped<ICountrySpecificQueueRepository, CountrySpecificQueueRepository>()
                 .AddScoped<ICountrySpecificSettings, CountrySpecificSettings>()
-                .AddSingleton(sp => QueueDEFAULTConfiguration.FromMiddlewareConfiguration(sp.GetRequiredService<MiddlewareConfiguration>()))
+                .AddSingleton(sp =>
```
pablo1609, line 27: needs to be improve

queue/src/fiskaltrust.Middleware.Queue/Bootstrapper/LocalizedQueueBootStrapperFactory.cs
pablo1609: In the last commits I've added your suggestions

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/Factories/SignatureItemFactoryDEFAULT.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        protected static NumberFormatInfo CurrencyFormatter = new ()
+        {
+            NumberDecimalSeparator = ",",
+            NumberGroupSeparator = "",
+            CurrencyDecimalDigits = 2
+        };
+
+        public SignaturItem[] CreateSignatures(DailyClosingResponse dailyClosingResponse) //CreateSignatures then remove the other functions
```
StefanKert, line 24: Should we add some signatures here like SANDBOX and such? Maybe also a QR Code with www.fiskaltrust.eu. So at least those signatures that we append to the country specific once can also be added :)

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/Factories/SignatureItemFactoryDEFAULT.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class SignatureItemFactoryDEFAULT : SignatureItemFactory
+    {
+        public override long CountryBaseState => 000000000;
+
+        public SignatureItemFactoryDEFAULT() { 
+        }
+
+        protected static NumberFormatInfo CurrencyFormatter = new ()
+        {
+            NumberDecimalSeparator = ",",
```
StefanKert, line 19: Not sure about this property, but if we are talking about 10.0 we should probably use the international format for this and choose a .

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/MonthlyClosingReceiptCommand.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public MonthlyClosingReceiptCommand(ICountrySpecificSettings countrySpecificSettings)
+        {
+            _countrySpecificQueueRepository = countrySpecificSettings.CountrySpecificQueueRepository;
+            _countryBaseState = countrySpecificSettings.CountryBaseState;
+        }
+
+        protected override async Task<string> GetCashboxIdentificationAsync(Guid ftQueueId)
+        {
+            var queueIt = await _countrySpecificQueueRepository.GetQueueAsync(ftQueueId).ConfigureAwait(false);
```
StefanKert, line 25: very small thing, but I guess this should be queueDefault

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/OutOfOperationReceiptCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+        public override Task<bool> ReceiptNeedsReprocessing(ftQueue queue, ReceiptRequest request, ftQueueItem queueItem) => Task.FromResult(false);
+
+        protected override Task<(ftActionJournal, SignaturItem)> DeactivateSCUAsync(ftQueue queue, ReceiptRequest request, ftQueueItem queueItem)
+        {
+            return Task.FromResult((new ftActionJournal {}, new SignaturItem {}));
+        }
+
+        protected override async Task<string> GetCashboxIdentificationAsync(Guid ftQueueId)
+        {
+            var queueIt = await _countrySpecificQueueRepository.GetQueueAsync(ftQueueId).ConfigureAwait(false);
```
StefanKert, line 37: queueDefault :)

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/PosReceiptCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            return Task.FromResult(new RequestCommandResponse
+            {
+                ReceiptResponse = new ReceiptResponse
+                {
+                    cbTerminalID = request.cbTerminalID,
+                    cbReceiptReference = request.cbReceiptReference,
+                    ftCashBoxIdentification = $"ft{queueItem.ftQueueRow}",
+                    ftQueueID = request.ftQueueID,
+                    ftQueueItemID = $"ft{queueItem.ftQueueItemId}",
```
StefanKert, line 45: ftQueueItemId should be the GUID of the queueitem so no ft prefix needed

(https://github.com/fiskaltrust/middleware/blob/29b8170ffbfe64a528337a5fdcc754d37a6653e2/queue/src/fiskaltrust.Middleware.Localization.QueueDE/RequestCommands/RequestCommand.cs#L168 for comparision)

```suggestion
                    ftQueueItemID = $"queueItem.ftQueueItemId",
```

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/PosReceiptCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public override Task<RequestCommandResponse> ExecuteAsync(ftQueue queue, ReceiptRequest request, ftQueueItem queueItem, bool isBeingResent = false)
+        {
+            return Task.FromResult(new RequestCommandResponse
+            {
+                ReceiptResponse = new ReceiptResponse
+                {
+                    cbTerminalID = request.cbTerminalID,
+                    cbReceiptReference = request.cbReceiptReference,
+                    ftCashBoxIdentification = $"ft{queueItem.ftQueueRow}",
```
StefanKert, line 43: ftCashBoxIdentification is a field from the countryspecificqueue :)

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/PosReceiptCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+                    ftQueueID = request.ftQueueID,
```
StefanKert, line 44: Should be taken from the queueItem since we need to closely relate ftReceiptJournal with ftQueueItem

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/PosReceiptCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            _countryspecificSettings = countrySpecificSettings;
+            _countrySpecificQueueRepository = countrySpecificSettings.CountrySpecificQueueRepository;
+            _countryBaseState = countrySpecificSettings.CountryBaseState;
+            _configurationRepository = configurationRepository;
+        }
+
+        public override Task<RequestCommandResponse> ExecuteAsync(ftQueue queue, ReceiptRequest request, ftQueueItem queueItem, bool isBeingResent = false)
+        {
+            return Task.FromResult(new RequestCommandResponse
```
StefanKert, line 37: So IMO we should change this and adapt it to be closer to what we have in germany e.g.

https://github.com/fiskaltrust/middleware/blob/29b8170ffbfe64a528337a5fdcc754d37a6653e2/queue/src/fiskaltrust.Middleware.Localization.QueueDE/RequestCommands/RequestCommand.cs#L168

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/YearlyClosingReceiptCommand.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public YearlyClosingReceiptCommand(ICountrySpecificSettings countryspecificSettings)
+        {
+            _countrySpecificQueueRepository = countryspecificSettings.CountrySpecificQueueRepository;
+            _countryBaseState = countryspecificSettings.CountryBaseState;
+        }
+
+        protected override async Task<string> GetCashboxIdentificationAsync(Guid ftQueueId)
+        {
+            var queueIt = await _countrySpecificQueueRepository.GetQueueAsync(ftQueueId).ConfigureAwait(false);
```
StefanKert, line 24: queueDefault

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/OutOfOperationReceiptCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+using fiskaltrust.Middleware.Contracts.Constants;
+using fiskaltrust.Middleware.Contracts.Repositories;
+using fiskaltrust.Middleware.Localization.QueueDEFAULT.Factories;
+using fiskaltrust.storage.serialization.DE.V0;
+using fiskaltrust.storage.V0;
+using Newtonsoft.Json;
+
+namespace fiskaltrust.Middleware.Localization.QueueDEFAULT.RequestCommands
+{
+    public class OutOfOperationReceiptCommand : Contracts.RequestCommands.OutOfOperationReceiptCommand
```
StefanKert, line 13: IMO this class hierarchy is very confusing. It makes it hard to understand the process and also hard to review. Currently I am not sure if we forget to set StopMoment or if it is set in the BaseCommand 🤔 Currently have no better idea for solving that but leaving this comment so that if anyone has an idea we can discuss that :)

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/Dummy/IMiddlewareJournalRepository.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Threading.Tasks;
+using fiskaltrust.storage.V0;
+
+namespace fiskaltrust.Middleware.Localization.QueueDEFAULT.Dummy
+{
+    public interface IMiddlewareJournalRepository
+    {
+        Task<ftJournalIT> GetByQueueItemId(Guid queueItemId);
```
StefanKert, line 9: We shouldn't return ftJournalIT here IMO :)

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/JournalProcessorDEFAULT.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public JournalProcessorDEFAULT(
+            ILogger<JournalProcessorDEFAULT> logger)
+        {
+            _logger = logger;
+        }
+
+        public IAsyncEnumerable<JournalResponse> ProcessAsync(JournalRequest request)
+        {
+            throw new NotImplementedException();
```
StefanKert, line 21: What are we usually returning when a JournalType is not processable?

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/Repositories/CountrySpecificQueueRepository.cs
```diff
@@ -0,0 +1,23 @@
...
+        private readonly IConfigurationRepository _configurationRepository;
+
+        public CountrySpecificQueueRepository(IConfigurationRepository configurationRepository)
+        {
+            _configurationRepository = configurationRepository;
+        }
+
+        public Task<ICountrySpecificQueue> GetQueueAsync(Guid queueId) => Task.FromResult<ICountrySpecificQueue>(new CountryDefaultQueue());
+
+        public Task InsertOrUpdateQueueAsync(ICountrySpecificQueue countrySpecificQueue) => Task.CompletedTask;
```
StefanKert, line 20: This means that we will never increase the counters that we have right? Also FailedCounters are usually set in the countryspecific queue. Is this something that we should look out for?

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/DailyClosingReceiptCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            _countrySpecificQueueRepository = countrySpecificSettings.CountrySpecificQueueRepository;
+            _signatureItemFactoryDefault = signatureItemFactoryDefault;
+            _countryspecificSettings = countrySpecificSettings;
+            _countryBaseState = countrySpecificSettings.CountryBaseState;
+        }
+
+        protected override async Task<string> GetCashboxIdentificationAsync(Guid ftQueueId)
+        {
+            var queueIt = await _countrySpecificQueueRepository.GetQueueAsync(ftQueueId).ConfigureAwait(false);
```
StefanKert, line 32: queueDefault

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/ZeroReceiptCommand.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace fiskaltrust.Middleware.Localization.QueueDEFAULT.RequestCommands
+{
+    public class ZeroReceiptCommand : Contracts.RequestCommands.ZeroReceiptCommand
+    {
+        public ZeroReceiptCommand(ICountrySpecificSettings countryspecificSettings, IMiddlewareQueueItemRepository queueItemRepository, IRequestCommandFactory requestCommandFactory, ILogger<RequestCommand> logger, IActionJournalRepository actionJournalRepository) :
+            base(countryspecificSettings, queueItemRepository, requestCommandFactory, logger, actionJournalRepository)
+        {
+        }
+    }
```
StefanKert, line 16: So not sure if the logic for recovering is in the baseclass but just leaving this here inc ase we forgot it :)

queue/src/fiskaltrust.Middleware.Queue/ReceiptRequestHelper.cs
```diff
@@ -178,7 +178,10 @@ public static string GetCountry(ReceiptRequest data)
             {
                 0x4445000000000000 => "DE",
                 0x4652000000000000 => "FR",
-                _ => "AT",
+                0x4154000000000000 => "AT",
+                0x4D45000000000000 => "ME",
+                0x4954000000000000 => "IT",
```
StefanKert, line 183: Does this mean that we didn't really have support for IT in yet? 

Also the fallback with AT is something that we should take a closer look at. Do you think having the DEFAULT fallback could cause troubles @TSchmiedlechner ?

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/PosReceiptCommand.cs
volllly: You mean don't create the `ReceiptResponse` inline but introduce a `CreateReceiptResponse` method that handles that?

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/PosReceiptCommand.cs
volllly: I'm not sure I understand 🤔 you mean just do `ftQueueID = queueItem.ftQueueID` instead?

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/OutOfOperationReceiptCommand.cs
volllly: The idea was that everything which is common to all markets (maybe you can think of it as everything that would be documented in the general part of the interface-docs) is done in the base classes.

It does get hard to review. The idea is that if we give the abstract methods in the base class meaningful names it should be clear what still has to be done by the country specific command.

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/JournalProcessorDEFAULT.cs
volllly: In Italy we also just throw ATM.

I Germany we return this if the case does not match https://github.com/fiskaltrust/middleware/blob/main/queue/src/fiskaltrust.Middleware.Localization.QueueDE/JournalProcessorDE.cs#L104-L111.

I think returning something similar makes sense but I would only do that on case `0x...0000` and still trow (with a nice error log) if the case does not match.

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/Repositories/CountrySpecificQueueRepository.cs
volllly: The thing is that we don't have these counters because we don't have a default countryspecific queue because we did not want to add default stuff to the storage. Since we also don't sign in the default queue we don't have failed counters 🤔 

BTW this PR is for now only the default queue and in the next step we will add comments and more dummy code to make it the new market template. then we will add explanations and todos for all of that.

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/ZeroReceiptCommand.cs
volllly: It is in the base class 👌 we have these `ICountrySpecificSettings` which contains a field `ResendFailedReceipts` that determines if failed receipts are resent.

https://github.com/fiskaltrust/middleware/blob/2c285afbc67667ab87738c5683114af5f6478d89/queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/Constants/CountrySpecificSettings.cs

As mentioned above receipts cannot fail in the default queue since we don't sign 😁

queue/src/fiskaltrust.Middleware.Queue/ReceiptRequestHelper.cs
volllly: Yes. we should definitely add IT in the main branch as well before we release the next version 😅 This method is only used here https://github.com/fiskaltrust/middleware/blob/2c285afbc67667ab87738c5683114af5f6478d89/queue/src/fiskaltrust.Middleware.Queue/SignProcessor.cs#L137 and `ftQueueItem.country` is not read anywhere in the MW (maybe in the portal?)

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/JournalProcessorDEFAULT.cs
volllly: Maybe we can extract that to a `JournalCommand` like we do with the receipt commands 🤔

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/Factories/SignatureItemFactoryDEFAULT.cs
volllly: The `S A N D B O X` signature is added by the `Queue` https://github.com/fiskaltrust/middleware/blob/main/queue/src/fiskaltrust.Middleware.Queue/SignatureFactory.cs. A default QR code is a nice idea 🙏

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/QueueDEFAULTConfiguration.cs
```diff
@@ -6,6 +6,8 @@ namespace fiskaltrust.Middleware.Localization.QueueDEFAULT
     public class QueueDEFAULTConfiguration
     {
         public bool Sandbox { get; set; } = true;
+        public bool FlagOptionalSignatures { get; set; }
```
volllly, line 9: This parameter we don't need in the Default queue. There are no optional signatures here 😅

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/Factories/SignatureItemFactoryDEFAULT.cs
```diff
@@ -16,17 +14,26 @@ public class SignatureItemFactoryDEFAULT : SignatureItemFactory
...
-
-        public SignaturItem[] CreateSignatures(DailyClosingResponse dailyClosingResponse) //CreateSignatures then remove the other functions
+        
+        public List<SignaturItem> GetSignaturesForPosReceiptTransaction()
         {
-            return new SignaturItem[]
+            var signatures = new List<SignaturItem>
             {
-                
+                CreateQrCodeSignature("www.fiskaltrust.eu", "<data-for-qr-code>")
```
volllly, line 26: @StefanKert what do you think we should put as qr-code data?

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/Factories/SignatureItemFactoryDEFAULT.cs
```diff
@@ -16,17 +14,26 @@ public class SignatureItemFactoryDEFAULT : SignatureItemFactory
...
             };
+
+            return signatures;
         }
+
+        public SignaturItem CreateQrCodeSignature(string caption, string data) => new()
+        {
+            Caption = caption,
+            ftSignatureFormat = (long)SignaturItem.Formats.QR_Code,
+            Data = data
```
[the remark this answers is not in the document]
pablo1609: I've set here ` ftSignatureType` because in definition of `SignaturItem` type is called here like this

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/OutOfOperationReceiptCommand.cs
StefanKert: Makes sense :)

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/RequestCommands/PosReceiptCommand.cs
StefanKert: Well no  not really. More or less we should use the one from the base class and augement it if necessary. :)

queue/src/fiskaltrust.Middleware.Queue/ReceiptRequestHelper.cs
StefanKert: Okay so this would be some sort of breaking change if we add it this way. All folks that are sending receiptcases that are "wrong" have gotten AT so far (which is not really correct but fair enough). In the future they would get "DEFAULT". I guess we should probably leave AT as defaut. Probably @TSchmiedlechner can add his two cents here since he has been more involved in implementations than me.

queue/src/fiskaltrust.Middleware.Localization.QueueDEFAULT/Factories/SignatureItemFactoryDEFAULT.cs
```diff
@@ -0,0 +1,40 @@
...
+            NumberDecimalSeparator = ".",
+            NumberGroupSeparator = "",
+            CurrencyDecimalDigits = 2
+        };
+        
+        public List<SignaturItem> GetSignaturesForPosReceiptTransaction()
+        {
+            var signatures = new List<SignaturItem>
+            {
+                CreateQrCodeSignature("www.fiskaltrust.eu", "<ftCashBoxIdentification>_<ftReceiptIdentification>_<SumOfPayItems>_<responseHashOfThePreviousQueueItem>")
```
StefanKert, line 26: IMO we will have to replace this with actual values, or is this done in the caller?

pablo1609: Need to replace it

forsthug (approved):

PR #173, opened 2023-07-06
Fiskaly tx count fix

### LINKED ISSUE #170: Fiskaly number of transactions maxes at 100
The current implementation of the FiskalyCertified SCU doesn't report the actual count of open transactions if there are more than 100. This is due to the fact that the Fiskaly API uses pagination and we are basing the count on the retruned active transactions.

Since the case of > 100 open transactions is not really likely we shouldn't have to download all transactions to the scu because that could delay the startup.

What we should fix though is the count that is being reported. The TSS endpoint offers the actual count of open transactions (https://developer.fiskaly.com/api/kassensichv/v2#operation/retrieveTss) so we can use that instead of the current implementation

paulcristiann:
- [X] You've read the [Contributor Guide](https://github.com/fiskaltrust/.github/blob/main/CONTRIBUTING.md) and [Code of Conduct](https://github.com/fiskaltrust/.github/blob/main/CODE_OF_CONDUCT.md).
- [X] You've included unit or integration tests for your change, where applicable.
- [X] You've included inline docs for your change, where applicable.
- [X] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

Changed the counter of the active transactions to be read directly from the API response

## Description

The active transactions counter is now parsed from the API response and reported correctly

Fixes #170

### DIFF
--- a/scu-de/src/fiskaltrust.Middleware.SCU.DE.FiskalyCertified/FiskalySCU.cs
+++ b/scu-de/src/fiskaltrust.Middleware.SCU.DE.FiskalyCertified/FiskalySCU.cs
@@ -148,7 +148,7 @@ namespace fiskaltrust.Middleware.SCU.DE.FiskalyCertified
                 return new TseInfo
                 {
                     CurrentNumberOfClients = clientDto.Where(x => x.State.Equals("REGISTERED")).ToList().Count,
-                    CurrentNumberOfStartedTransactions = startedTransactions.Count(),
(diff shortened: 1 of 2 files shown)


paulcristiann:
recheck

paulcristiann:

paulcristiann:
recheck

volllly (approved):

PR #192, opened 2023-08-01
RFC: France copy receipt

volllly:
Adding a `ftJournalFRCopyReceipts` table in the FR Queue where we store copy receipts.

[Rendered](https://github.com/fiskaltrust/middleware/blob/rfc/france-copy-receipt/rfcs/0192-fr-copy-receipt-performance.md)

- [x] Remove tmp code changes form the repo
- [x] Finish proposal
- [x] Answer unresolved/open questions


rfcs/192-fr-copy-receipt-performance.md
```diff
@@ -0,0 +1,177 @@
...
+For the PosCreator they are no implementation changes needed, the CopyReceipts still work like they did before.
+
+## New local queues
+
+New Queues don't need to take anything into account.
+They will work out of the box.
+
+## Updating local queues
+
+The new table will be created on start up after updating the queue.
```
StefanKert, line 30: We should probably do some checks how long this can take. Maybe we can perform this also async in the background after the queue has started and becomes available?

rfcs/192-fr-copy-receipt-performance.md
```diff
@@ -0,0 +1,177 @@
...
+
+The data will be migrated from the `ftJournalFR` table to the new `ftJournalFRCopyReceipts` table.
```
StefanKert, line 32: Migrating is probably not the correct word. We are creating lookup references / entries, but we'll keep the entries in the ftJournalFR table.

rfcs/192-fr-copy-receipt-performance.md
```diff
@@ -0,0 +1,177 @@
...
+This will take a similar timeframe as it takes to process a CopyReceipt with the old implementation.
+
+## SignatureCloud
+
+The SignatureCloud will be updated as well.
```
StefanKert, line 37: @Bluette-C-Riviere this is more of a certification question IMO. What would be the impact if we update all signaturcloud customers? Can you probably figure out how many of them are affected in if that makes sense?

rfcs/192-fr-copy-receipt-performance.md
```diff
@@ -0,0 +1,177 @@
...
+# Drawbacks
+
+This solution involves a new table and we need to populate it with the data from the `ftJournalFR` table before we can use the new implementation.
+This brings some problems with it especially in the SignatureCloud.
+
+This update version of the middleware might need to be certified again.
+
+# Rationale and alternatives
+
+An alternative way of doing this would be to keep the current implementation and improve it.
```
StefanKert, line 129: @TSchmiedlechner @steininger even though this option wouldn't necessarily solve all of our problems it would be very little effort and could already improve performance hugely. Curious about your opinion

rfcs/192-fr-copy-receipt-performance.md
volllly: Hm we could maybe do that but we'll have to wait for the population of the `ftJournalFRCopyReceipts` to be completed before processing a copy receipt... I'm not sure if that complexity is worth it

rfcs/192-fr-copy-receipt-performance.md
```diff
@@ -0,0 +1,177 @@
...
+    return count;
+}
+```
+
+## SignatureCloud
+
+We will setup a new SignatureCloud environment which uses the new implementation and keep using the old one in the beginning.
+We will have a load balancer in front of the SignatureCloud which will route the requests to the correct environment.
+
+When the PosCreator is ready to use the new implementation we will switch the load balancer to the new environment.
```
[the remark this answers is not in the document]
volllly: I think that's only possible if the signaturecloud does the table population automatically. (see Unresolved questions)

rfcs/192-fr-copy-receipt-performance.md
volllly: I think if it's possible it would be a nice intermediate solution that would relive some pressure and give us more time to deal with the situation.

rfcs/192-fr-copy-receipt-performance.md
Bluette-C-Riviere: We have 4 customers/dealers using or having operating using the ChaineCloud

only 2 would need a new certificate, a third one will not be impacted because the certificate should be given mid-october.
The new certificate will be edited by the LNE after going through the change management process with have with them.

rfcs/192-fr-copy-receipt-performance.md
TSchmiedlechner: Interesting idea, why not :) I'd make sure to not "certify" the intermediate solution though, so we'd maybe have multiple RCs.

volllly: I've just thought some more about that and it would not really help (in this form) since the first copyreceipt of a receipt would always iterate through all receipts and then return 0... we would also have to check for normal receipts and check their receiptreferences and stop with count 0 if we find one. That would maybe change the behaviour of the implementation in certain edge cases.

Bluette-C-Riviere: As far as certification goes,  if i understood correctly we will use self-testation for the RCs

StefanKert: > Interesting idea, why not :) I'd make sure to not "certify" the intermediate solution though, so we'd maybe have multiple RCs.

So as far as I understood @Bluette-C-Riviere we will anyways have to choose the self testation part regardless of what we do. So to show some progress we could implement this suggested solution in rc1 roll this out for sqlite and see if it solves the most urgent issues. We then could keep on looking for the approach described elarier that involves a lookup table

rfcs/192-fr-copy-receipt-performance.md
```diff
@@ -0,0 +1,186 @@
...
+# Rationale and alternatives
+
+An alternative way of doing this would be to keep the current implementation and improve it.
+
+One way to do that would be to not count every CopyReceipt in the `ftJournalFR` table but only find the latest one and increase its count by one.
+
+If we find a non copy receipt with the same `cbReceiptReference` as the `cbPreviousReceiptReference` we can return 0 directly.
+
+```cs
+private async Task<int> GetCountOfExistingCopiesAsync(string cbPreviousReceiptReference)
```
StefanKert, line 136: So IMO we should be good if we start including this fix for a first `rc`. Based on how this fix performs we can then decide how we build up the table ( maybe in process, or at the beginning).

volllly:
> One thing that we can also consider as potential solution is using the RawData that is stored in the JournalFR. For CopyReceipts we are only considering items with Type 'C'. These items include the JWT token the data for those 'C' entries has always the same format:

This suggestion has been added to the RFC and released to sandbox as the first `rc` version in `fiskaltrust.service.sqlite.1.2.23219.65461-rc1`.

rfcs/0192-fr-copy-receipt-performance.md
```diff
@@ -0,0 +1,218 @@
...
+    var query = "SELECT count(*) FROM ftJournalFRCopyPayload WHERE CopiedReceiptReference = @cbPreviousReceiptReference";
+    return await DbConnection.QuerySingleAsync<int>(query, new { cbPreviousReceiptReference });
+}
+```
+
+## Initialization
+
+The new table will be created with the usual migration process.
+
+On start up of the Queue (in the 1.2 `workerFR.cs`) we check if the table contains any entries and if not we will populate it with the old data from the `ftJournalFR` table.
```
TSchmiedlechner, line 84: I guess it would also check if the queue is started and especially has already processed copy receipts (maybe with the `CNumerator` of the QueueFR?), to not execute this multiple times whenever a Queue is started that doesn't contain copy receipts

rfcs/0192-fr-copy-receipt-performance.md
```diff
@@ -0,0 +1,218 @@
...
+
+The SignatureCloud will be updated as well.
+The new version will not automatically be used since this would mean a recertification of the PosCreator.
+Instead we will time the switch to the new version accordingly with the PosCreators.
+
+# Reference-level explanation
+
+We will add a migration which adds the new table `ftJournalFRCopyPayload`.
+
+This Table will contain the following columns:
```
TSchmiedlechner, line 44: Might make sense to also add the `CNumerator` here and sort by that DESC, right? By doing this, we'll only have to load a single row and can
a) make the lookup faster in Azure Storage by sorting DESC by this
b) make the lookup faster in SQL databases by putting an index on this field.

rfcs/0192-fr-copy-receipt-performance.md
```diff
@@ -0,0 +1,218 @@
...
+  And do we have clients which have a lot of CopyReceipts and need copies of older receipts where this would lean towards worst case performance?
+
+# Unresolved questions
+
+## Initialization
+
+Can we populate the table asynchronously?
+This would mean we could start the Queue (and process non CopyReceipts) without waiting for the population of the `ftJournalFRCopyPayload` table to finish.
+
+However this would mean we would have to either fail CopyReceipts until the process of population is complete or wait for the population to finish before we start processing the CopyReceipt.
```
TSchmiedlechner, line 202: We'd need to test how long this takes i guess, IMO everything up to 25sec startup time (one-shot) would be okayish for existing implementations (given that the max startup time of a Windows service is 30 seconds)

TSchmiedlechner (approved):

PR #208, opened 2023-09-27
Use DATABASE() method instead of templating in mysql migrations

volllly:


queue/src/fiskaltrust.Middleware.Storage.SQLite/Migrations/012_ftJournalFRCopyPayload.sqlite3
pawelvds, line 1: Thanks for updated it too 🥇

pawelvds (approved):

PR #209, opened 2023-10-13
Remove sync calls from async code

volllly:


queue/src/fiskaltrust.Middleware.Localization.QueueDE/JournalProcessorDE.cs
```diff
@@ -103,7 +103,7 @@ await foreach (var value in ProcessTarExportFromDatabaseAsync(request).Configure
             {
                 var result = new
                 {
-                    QueueDEList = _configurationRepository.GetQueueDEListAsync().Result
+                    QueueDEList = await _configurationRepository.GetQueueDEListAsync()
```
StefanKert, line 106: Do we now why this has been done that way? This looks awful

queue/src/fiskaltrust.Middleware.Localization.QueueFR/RequestCommands/ZeroReceiptCommand.cs
```diff
@@ -81,7 +81,7 @@ private List<ftActionJournal> ResetFailedMode(ReceiptRequest request, ReceiptRes
                 var fromReceipt = "#";
                 try
                 {
-                    var fromQueueItem = _queueItemRepository.GetAsync(queueFR.UsedFailedQueueItemId.Value).Result;
+                    var fromQueueItem = await _queueItemRepository.GetAsync(queueFR.UsedFailedQueueItemId.Value);
```
StefanKert, line 84: Do we need to ConfigureAwait(false)

queue/src/fiskaltrust.Middleware.Localization.QueueFR/RequestCommands/ZeroReceiptCommand.cs
```diff
@@ -33,7 +33,7 @@ public override async Task<(ReceiptResponse receiptResponse, ftJournalFR journal
                 var (response, journalFR) = CreateTrainingReceiptResponse(queue, queueFR, request, queueItem, totals, signaturCreationUnitFR);
                 response.ftSignatures = response.ftSignatures.Extend(await GetPendingMessageSignatures(queueFR, request));
 
-                var actionJournals = ResetFailedMode(request, response, queue, queueFR, queueItem);
+                var actionJournals = await ResetFailedMode(request, response, queue, queueFR, queueItem);
```
StefanKert, line 36: ResetFailedModeAsync. Also we probably need ConfigureAwait(false)?

queue/src/fiskaltrust.Middleware.Localization.QueueFR/RequestCommands/ZeroReceiptCommand.cs
```diff
@@ -45,7 +45,7 @@ public override async Task<(ReceiptResponse receiptResponse, ftJournalFR journal
 
                 var payload = PayloadFactory.GetGrandTotalPayload(request, response, queueFR, signaturCreationUnitFR, queueFR.GLastHash);
 
-                var actionJournals = ResetFailedMode(request, response, queue, queueFR, queueItem);
+                var actionJournals = await ResetFailedMode(request, response, queue, queueFR, queueItem);
```
StefanKert, line 48: Also we probably need ConfigureAwait(false)?

queue/src/fiskaltrust.Middleware.Localization.QueueME/QueueMeBootstrapper.cs
```diff
@@ -18,18 +20,16 @@ public void ConfigureServices(IServiceCollection services)
                 .AddScoped<IMarketSpecificSignProcessor, SignProcessorME>()
                 .AddScoped<IMarketSpecificJournalProcessor, JournalProcessorME>()
                 .AddSingleton(sp => QueueMEConfiguration.FromMiddlewareConfiguration(sp.GetRequiredService<MiddlewareConfiguration>()))
-                .AddSingleton<IMESSCDProvider>(sp =>
-                {
-                    var sscdProvider = new MESSCDProvider(
-                        sp.GetRequiredService<IClientFactory<IMESSCD>>(),
-                        sp.GetRequiredService<MiddlewareConfiguration>());
-                    sscdProvider.RegisterCurrentScuAsync().Wait();
```
StefanKert, line 26: I guess this is now done internally right?

queue/src/fiskaltrust.Middleware.Localization.QueueDE/QueueDEBootstrapper.cs
```diff
@@ -50,9 +57,20 @@ public void ConfigureServices(IServiceCollection services)
...
+
+        public Task<Func<IServiceProvider, Task>> ConfigureServicesAsync(IServiceCollection services)
+        {
+            var _ = ConfigureServicesInner(services)
+                .AddSingleton<IDESSCDProvider, DESSCDProvider>()
+                .AddSingleton<ITarFileCleanupService, TarFileCleanupService>();
+
+            return Task.FromResult<Func<IServiceProvider, Task>>(async (IServiceProvider services) =>
+            {
+                await services.GetRequiredService<IDESSCDProvider>().RegisterCurrentScuAsync();
```
StefanKert, line 71: So I guess we could discuss this in another issue, but IMO we should not register the scu in our configure services method.. also probably not during resolution of the instance, but more or less during the first creation. By doing that we can probably get rid of some tasking things

queue/src/fiskaltrust.Middleware.Localization.QueueDE/QueueDEBootstrapper.cs
```diff
@@ -50,9 +57,20 @@ public void ConfigureServices(IServiceCollection services)
...
+                await services.GetRequiredService<ITarFileCleanupService>().CleanupAllTarFilesAsync();
```
StefanKert, line 72: This seems to be.. yeah.. I don't know. should we put that somewhere else and not during the startup? Can we create a background thread that performs this ? maybe another issue, but this has caused major pain in the past

queue/src/fiskaltrust.Middleware.Storage.SQLite/SQLiteStorageBootstrapper.cs
```diff
@@ -49,6 +49,12 @@ public void ConfigureStorageServices(IServiceCollection serviceCollection)
             AddRepositories(serviceCollection);
         }
 
+        public async Task ConfigureStorageServicesAsync(IServiceCollection serviceCollection)
+        {
+            await InitAsync(_queueId, _configuration, _logger);
```
StefanKert, line 54: Welp thats another thing. So I am not sure but IMO we should split up the initialization of our objects from the initialization of the database.

StefanKert:
So generally speaking I think that the API that we have added here is okay. Still I think that we should probably rethink how we initialize the Middleware in a general way. Right now we are doing lots of steps that take time during the startup of the launcher which affects a few processes IMO. Especially during restarts of machines this puts a lot of pressure on the computer so we should probably reconsider which things we perform during this hot phase.

queue/src/fiskaltrust.Middleware.Localization.QueueDE/JournalProcessorDE.cs
volllly: This is something that I'll test but I hope there was no reason 😅

TSchmiedlechner:
I agree with @StefanKert on one hand, but on the other hand it might be better to do this before the first receipt arrives and not _during_ that - if the first receipt after startup takes multiple seconds, this would be worse than an equally slow startup IMO

PR #227, opened 2023-11-23
Enhanced the transaction already open exception text

saschaknapp:
In the daily business there are cases from time to time where the SCU throws the exception "Transaction already finished...". This exception is caused by the transaction being closed in the TSE but still being open in the middleware openTransaction table. This can easily fixed by adding the flag to the daily-closing. The middleware will then close the open transaction in its database.

I added the fix to the exception text, so that customers and the fiskaltrust support know how to fix the issue immediately.


volllly (approved):

saschaknapp:

saschaknapp:
recheck

PR #229, opened 2023-12-05
Update masterdate-add actionjournal- add dailyclosingnumber when tse …

LeilaZarban:
…is not reachable

{Summary of the changes}

## Description

{Detail}

Fixes #{issue number}

### DIFF
--- a/queue/src/fiskaltrust.Middleware.Localization.QueueDE/RequestCommands/DailyClosingReceiptCommand.cs
+++ b/queue/src/fiskaltrust.Middleware.Localization.QueueDE/RequestCommands/DailyClosingReceiptCommand.cs
@@ -22,7 +22,7 @@ namespace fiskaltrust.Middleware.Localization.QueueDE.RequestCommands
     {
         public override string ReceiptName => "Daily-closing receipt";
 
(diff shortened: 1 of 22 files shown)


queue/src/fiskaltrust.Middleware.Localization.QueueDE/RequestCommands/DailyClosingReceiptCommand.cs
```diff
@@ -126,13 +115,65 @@ public override async Task<RequestCommandResponse> ExecuteAsync(ftQueue queue, f
...
+                        Moment = DateTime.UtcNow,
+                        TimeStamp = DateTime.UtcNow.Ticks,
+                        Priority = -1,
+                        DataJson = JsonConvert.SerializeObject(new
+                        {
+                            ftReceiptNumerator = queue.ftReceiptNumerator + 1,
+                            masterDataChanged = masterDataChanged,
+                            closingNumber = queueDE.DailyClosingNumber
+                        })
+                    });
```
volllly, line 139: I think we could reuse the `CreateClosingActionJournals` method thats also used above for creating the actionJournal.

We'd need to change the method to accept a `ulong? transactionNumber` and I think then we can use it like that 🤔 we'd then explicitly have `"transactionNumber": null` in the actionJournal as an indicator that it was not signed which could be practical when a human is reading it ^^

```suggestion
                actionJournals.Add(
                  CreateClosingActionJournals(queueItem, queue, null, masterDataChanged, $"{message} However TSE was not reachable.", type, queueDE.DailyClosingNumber));
```

LeilaZarban: Its done

PR #285, opened 2024-05-14
Fix ftReceiptTotal

LeilaZarban:

{Summary of the changes}

## Description

{Detail}

Fixes #{issue number}

### DIFF
--- a/queue/src/fiskaltrust.Middleware.Queue/SignProcessor.cs
+++ b/queue/src/fiskaltrust.Middleware.Queue/SignProcessor.cs
@@ -351,7 +351,7 @@ namespace fiskaltrust.Middleware.Queue
             }
             else
             {
-                receiptjournal.ftReceiptTotal = (receiptrequest?.cbChargeItems?.Sum(ci => ci.Amount)).GetValueOrDefault();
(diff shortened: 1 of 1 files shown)


queue/src/fiskaltrust.Middleware.Queue/SignProcessor.cs
```diff
@@ -351,7 +351,7 @@ public async Task<ftReceiptJournal> CreateReceiptJournalAsync(ftQueue queue, ftQ
             }
             else
             {
-                receiptjournal.ftReceiptTotal = (receiptrequest?.cbChargeItems?.Sum(ci => ci.Amount)).GetValueOrDefault();
+                receiptjournal.ftReceiptTotal = (receiptrequest?.cbChargeItems?.Sum(ci => Math.Sign(ci.Quantity) * ci.Amount)).GetValueOrDefault();
```
StefanKert, line 354: ftReceiptTotal should always be the straight sum of the Amounts. This is due to the definition of how we have specified usage of Amount / chargeitems in the interface docs. Taking the Quantity into the calculation will make the calculation wrong for a lot of cases so we must not change that.

PR #295, opened 2024-05-23
Fix flag calculation in IT

volllly:


pawelvds (approved):

queue/src/fiskaltrust.Middleware.Localization.QueueIT/Extensions/ReceiptRequestExtensions.cs
```diff
@@ -4,9 +4,9 @@ namespace fiskaltrust.Middleware.Localization.QueueIT.Extensions
 {
     public static class ReceiptRequestExtensions
     {
-        public static bool IsVoid(this ReceiptRequest receiptRequest) => (receiptRequest.ftReceiptCase & 0x0000_0000_000F_0000) == 0x0000_0000_0004_0000;
+        public static bool IsVoid(this ReceiptRequest receiptRequest) => (receiptRequest.ftReceiptCase & 0x0000_0000_0004_0000) > 0;
```
StefanKert, line 7: So that was my bad I guess. Can you add some unit tests to prevent my rookie mistakes in the future?

volllly: I can, but they would soon become obsolete after we've implemented #296 and moved all of the MW to use the source generated extensions

PR #306, opened 2024-06-18
Update AzureStorage table names

volllly:
We want to be backwards compatible with the 1.2 storage so we need to change the table names to match.


volllly:
Before the AzureStorage Queue is usable in FR release we'll need fiskaltrust/middleware#307

TSchmiedlechner (approved):

PR #399, opened 2025-02-18
Add SSCD fail ftState when Queue is in failed mode

TSchmiedlechner:
We should call `ProcessSSCDFailedReceipt` after performing the actual signing to also correctly set the ftState for the first receipt where the SSCD isn't reachable.


forsthug (approved):

volllly:
Fixes #377

PR #410, opened 2025-03-12
Fix sqlite mono dependencies

volllly:


.github/workflows/package.yml
```diff
@@ -39,6 +39,14 @@ jobs:
           certificate: '${{ secrets.SIGNING_CERTIFICATE }}'
           password: '${{ secrets.SIGNING_CERT_PASSWORD }}'
 
+      - run: |
+          dotnet publish --configuration Release --no-restore -f net461
+          dotnet publish --configuration Release --no-restore -f netstandard2.0
+          dotnet publish --configuration Release --no-restore -f netstandard2.1
+        if: inputs.pattern == 'fiskaltrust.Middleware.Queue.SQLite'
```
StefanKert, line 46: Do we need net8.0 / net6.0 too?

volllly: has been like that before,
launcher 2.0 is working aswell
so I guess not 🤔

volllly: something to worry about later

PR #433, opened 2025-05-05
Add compression to oversized azurestorage items

volllly:


queue/src/fiskaltrust.Middleware.Storage.AzureTableStorage/Mapping/Mapper.cs
```diff
@@ -50,16 +66,44 @@ public static string GetOversized(this TableEntity entity, string property)
             }
             else
             {
-                var marker = $"{property}_{OVERSIZED_MARKER}_{{0}}";
-                var reqSb = new StringBuilder();
-                for (var i = 0; entity.ContainsKey(string.Format(marker, i)); i++)
+                try
```
LeilaZarban, line 69: Can we avoid try/catch for flow control? I mean, detect the correct retrieval method up front, based on the stored format

PR #440, opened 2025-05-14
Fix nuspec

volllly:


volllly:
## Test plan

queue in all end2end tests

PR #451, opened 2025-06-24
Enrich Diagnostic Activities with additional data

volllly:


queue/src/fiskaltrust.Middleware.Localization.v2/SignProcessor.cs
```diff
@@ -112,11 +112,14 @@ public SignProcessor(
                 }
                 if (_isSandbox)
                 {
-                    receiptResponse.ftSignatures.Add(SignatureFactory.CreateSandboxSignature(_queueId));
+                    receiptResponse.AddSignatureItem(SignatureFactory.CreateSandboxSignature(_queueId));
                 }
 
                 await _queueStorageProvider.FinishQueueItem(queueItem, receiptResponse);
 
+                System.Diagnostics.Activity.Current?.AddTag("queue.ReceiptResponse.ftState", $"0x{receiptResponse.ftState:X}");
```
StefanKert, line 120: OH THIS IS HUGE

StefanKert: ReceiptResponse is always not null here right?

queue/src/fiskaltrust.Middleware.Queue/SignProcessor.cs
```diff
@@ -71,7 +71,14 @@ public async Task<ReceiptResponse> ProcessAsync(ReceiptRequest request)
 
                 var queue = await _configurationRepository.GetQueueAsync(_middlewareConfiguration.QueueId).ConfigureAwait(false);
 
-                return await InternalSign(queue, request).ConfigureAwait(false);
+                var response = await InternalSign(queue, request).ConfigureAwait(false);
+
+#if NET6_0_OR_GREATER
```
StefanKert, line 76: Can Activity be null?

queue/src/fiskaltrust.Middleware.Queue/SignProcessor.cs
```diff
@@ -178,7 +185,7 @@ private async Task<ReceiptResponse> InternalSign(ftQueue queue, ReceiptRequest d
                                 Data = e.ToString()
                             }
                         },
-                        ftState = (long) (((ulong) data.ftReceiptCase & 0xFFFF_0000_0000_0000) | 0x2000_EEEE_EEEE)
+                        ftState = (long)(((ulong)data.ftReceiptCase & 0xFFFF_0000_0000_0000) | 0x2000_EEEE_EEEE)
                     };
                 }
```
StefanKert, line 190: Should we create activity traces here too?

queue/src/fiskaltrust.Middleware.Localization.v2/SignProcessor.cs
```diff
@@ -156,7 +159,7 @@ private ReceiptResponse CreateReceiptResponse(ReceiptRequest receiptRequest, ftQ
             cbReceiptReference = receiptRequest.cbReceiptReference,
             ftCashBoxIdentification = _cashBoxIdentification,
             ftReceiptMoment = DateTime.UtcNow,
-            ftState = (State) ((ulong) receiptRequest.ftReceiptCase & 0xFFFF_F000_0000_0000),
+            ftState = (State)((ulong)receiptRequest.ftReceiptCase & 0xFFFF_F000_0000_0000),
```
StefanKert, line 162: Should we create activity traces here too?

queue/src/fiskaltrust.Middleware.QueueSynchronizer/fiskaltrust.Middleware.QueueSynchronizer.csproj
```diff
@@ -9,7 +9,8 @@
     </ItemGroup>
 
     <ItemGroup Condition="'$(TargetFramework)' != 'net461'">
-        <PackageReference Include="System.Threading.Channels" Version="7.0.0" />
+        <PackageReference Include="System.Threading.Tasks.Dataflow" Version="9.0.4" />
```
StefanKert, line 12: Do we need to exclude this for netstandard2.0?

volllly: If the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.dataflow?view=net-9.0#remarks) are correct this is only available as a nuget package (but it has no dependencies so we should not run into troubles)

queue/src/fiskaltrust.Middleware.Queue/SignProcessor.cs
volllly: you mean something like this?

`Activity.Current?.Source.StartActivity("InternalSign")`?

queue/src/fiskaltrust.Middleware.Queue/SignProcessor.cs
```diff
@@ -71,7 +71,14 @@ public async Task<ReceiptResponse> ProcessAsync(ReceiptRequest request)
 
                 var queue = await _configurationRepository.GetQueueAsync(_middlewareConfiguration.QueueId).ConfigureAwait(false);
 
-                return await InternalSign(queue, request).ConfigureAwait(false);
+                var response = await InternalSign(queue, request).ConfigureAwait(false);
+
+#if NET6_0_OR_GREATER
+                System.Diagnostics.Activity.Current?.AddTag("queue.ReceiptResponse.ftState", $"0x{response.ftState:X}");
```
StefanKert, line 77: response can be null in case of using the flag too lookup a receipt and not getting back anything :/

PR #465, opened 2025-07-09
Remove open transactions from database that don't exist on the TSE when automatically finishing them

volllly:
## Background

We store the open transactions the middleware knows about in the open transactions table.
On the daily closing we automatically finish these open transactions.

## Problem

When these transactions are already closed on the TSE (e.g. by another queue or because the SCU call failed at an inopportune time) this currently leads to the daily closing silently failing and returning an `ftState` with flag`0x2`.

## Solution

When finishing the open transaction during the daily closing fails check if the transaction still exists on the TSE.
If it does exist throw and handle that as before because something actually went wrong.

If it does not exist though it has to have been already finished (or never existed in the first hand) and can remove it from the open transaction table.

In this case though we can not return a signature for this finish transaction.


volllly:
## Test plan
configure two queues to use the same SCU.
create 1 implicit receipt on each queue.
create a daily closing with the close all open transactions flag on one queue.
create a daily closing on the other queue.
verify that the open transactions table in the database of the second queue is empty.

PR #472, opened 2025-07-16
Add journalES

volllly:


queue/src/fiskaltrust.Middleware.Storage.AzureTableStorage/Repositories/ES/AzureTableStorageJournalESRepository.cs
```diff
@@ -0,0 +1,88 @@
...
+using fiskaltrust.Middleware.Contracts.Repositories;
+using fiskaltrust.Middleware.Storage.AzureTableStorage.Mapping;
+using fiskaltrust.Middleware.Storage.AzureTableStorage.TableEntities;
+using fiskaltrust.storage.V0;
+
+namespace fiskaltrust.Middleware.Storage.AzureTableStorage.Repositories.ES
+{
+    public class AzureTableStorageJournalESRepository : BaseAzureTableStorageRepository<Guid, TableEntity, ftJournalES>, IJournalESRepository, IMiddlewareRepository<ftJournalES>
+    {
+        public AzureTableStorageJournalESRepository(QueueConfiguration queueConfig, TableServiceClient tableServiceClient)
```
StefanKert, line 14: Migration still needed

queue/src/fiskaltrust.Middleware.Localization.v2/JournalProcessor.cs
```diff
@@ -136,50 +129,48 @@ private List<T> GetConfigurationFromDictionary<T>(string key)
...
             {
-                yield return new JournalResponse
+                await foreach (var journal in asyncEnumerable)
                 {
-                    Chunk = buffer.Take(readAmount).ToList()
-                };
-                buffer = new byte[chunkSize];
+                    try
+                    {
+                        await response.Writer.WriteAsync(Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(journal)));
```
StefanKert, line 144: JsonConvert or JsonSerailizer?

queue/src/fiskaltrust.Middleware.Localization.v2/JournalProcessor.cs
```diff
@@ -136,50 +129,48 @@ private List<T> GetConfigurationFromDictionary<T>(string key)
...
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.LogError(ex, "An error occurred while writing journal data to the response\n.");
```
StefanKert, line 148: We should think about error handling here. We should make sure that no item is skipped, or if there is an error the export fails

storage/src/fiskaltrust.storage/Models/ftJournalES.cs
```diff
@@ -0,0 +1,22 @@
...
+        TicketBAI
+    }
+
+    public class ftJournalES
+    {
+        public Guid ftJournalESId { get; set; }
+        public long Number { get; set; }
+        public string RequestData { get; set; }
+        public string ResponseData { get; set; }
+        public string JournalType { get; set; }
```
StefanKert, line 17: Let's add a version here

queue/src/fiskaltrust.Middleware.Localization.v2/JournalProcessor.cs
volllly: I did not change that behavior. If there was a reason to use newtonsoft here it's still valid. (but IDK if there was a reason ^^)

queue/src/fiskaltrust.Middleware.Localization.QueueES/Models/MiddlewareState.cs
```diff
@@ -0,0 +1,54 @@
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using fiskaltrust.ifPOS.v2;
+
+namespace fiskaltrust.Middleware.Localization.QueueES.Models;
+
+public class MiddlewareState
```
volllly, line 8: These should in the future be defined in the interface or some common project. I already ran into problems during dev when changing it on one side only and not the other.

On the other hand there must never be changes in here that can break anything so we'll need robust regression/snapshot tests for that.

queue/src/fiskaltrust.Middleware.Localization.QueueES/Processors/ReceiptCommandProcessorES.cs
```diff
@@ -6,45 +6,108 @@
...
+                Number = response.ReceiptResponse.ftQueueRow,
+                Data = JsonSerializer.Serialize(responseStateData.ES!.GovernmentAPI),
+                JournalType = JournalESType.VeriFactu.ToString(),
+                ftQueueItemId = response.ReceiptResponse.ftQueueItemID,
+                ftQueueId = response.ReceiptResponse.ftQueueID,
+            });
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Error processing receipt");
```
volllly, line 108: We could first save the queueitem into the queue below and then throw if this fails but the receipt has been sent to verifactu in this case so I don't think that would be correct... should we add an error message as a signature?

queue/src/fiskaltrust.Middleware.Localization.QueueES/Models/MiddlewareState.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    [JsonPropertyName("ES")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public MiddlewareQueueESState? ES { get; set; }
+
+    [JsonExtensionData]
+    public Dictionary<string, JsonElement> ExtraData { get; set; } = new Dictionary<string, JsonElement>();
+}
```
StefanKert, line 16: The cbPreviousReceiptReference is also defined on this level right?

scu-es/src/fiskaltrust.Middleware.SCU.ES.VeriFactu/Soap/Client.cs
```diff
@@ -25,44 +26,44 @@ public Client(HttpClient httpClient)
...
+        {
+            Request = requestString,
+            Response = await response.Content.ReadAsStringAsync(),
+            Version = GovernmentAPISchemaVersion.V0
+        };
 
         if (!response.IsSuccessStatusCode || response.Content.Headers.ContentType?.MediaType == "text/html")
         {
-            return new Error.Http(response.StatusCode, await response.Content.ReadAsStringAsync());
+            return (new Error.Http(response.StatusCode, governmentAPI.Response), governmentAPI);
```
StefanKert, line 43: Should we also include the Request here? For debug reasons it could help having it available

queue/src/fiskaltrust.Middleware.Storage.AzureTableStorage/Repositories/ES/AzureTableStorageJournalESRepository.cs
```diff
@@ -0,0 +1,84 @@
...
+            if (src == null)
+            {
+                return null;
+            }
+
+            return new ftJournalES
+            {
+                ftJournalESId = src.GetGuid(nameof(ftJournalES.ftJournalESId)).GetValueOrDefault(),
+                ftQueueId = src.GetGuid(nameof(ftJournalES.ftQueueId)).GetValueOrDefault(),
+                ftQueueItemId = src.GetGuid(nameof(ftJournalES.ftQueueItemId)).GetValueOrDefault(),
```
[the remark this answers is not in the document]
volllly: I've just put the version inside of the GovernmentApi Data because that's the only data the scu is providing at the moment.. 🤔
That has the disadvantage though, that if we would change the `GovernmentApi` schema to have other fields then request and response we would need do a pre parsing pass to read it out first..
The advantage is that we don't need another field in the ftStateData that's inside our contract between queue and scu...
If we keep it inside the GovernmentApi Data only the SCU needs to know about it

queue/src/fiskaltrust.Middleware.Localization.QueueES/Processors/ReceiptCommandProcessorES.cs
```diff
@@ -6,45 +6,108 @@
...
+            }
+        };
+
         var response = await (await _essscd).ProcessReceiptAsync(new ProcessRequest
         {
             ReceiptRequest = request.ReceiptRequest,
             ReceiptResponse = request.ReceiptResponse,
-            PreviousReceiptRequest = previousQueueItem is null ? null : JsonSerializer.Deserialize<ReceiptRequest>(previousQueueItem.request)!, // handle null case?
-            PreviousReceiptResponse = previousQueueItem is null ? null : JsonSerializer.Deserialize<ReceiptResponse>(previousQueueItem.response)!,
+            ReferencedReceiptRequest = referencedQueueItem is null ? null : JsonSerializer.Deserialize<ReceiptRequest>(referencedQueueItem.request)!,
```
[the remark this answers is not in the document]
volllly: Will be done later when we have that in the localization v2

StefanKert (approved):

PR #490, opened 2025-08-08
Use Nerdbank.GitVersion for package versioning and releases

volllly:


storage/global.json
```diff
@@ -0,0 +1,6 @@
+{
+  "sdk": {
+    "version": "6.0.428",
```
volllly, line 3: **question:** we need that to get gitversion to work, right?

.github/workflows/command-version.yml
```diff
@@ -10,47 +9,30 @@ permissions:
...
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
 
       - name: Update version (nbgv prepare-release)
         run: |
-          # This will update version.json to the full release version without rc
-          nbgv prepare-release "" --nextVersion $(nbgv get-version -f json | jq -r '.Version')
-      
+          CURRENT_VERSION=$(nbgv get-version -f json | jq -r '.Version')
+          nbgv set-version "$CURRENT_VERSION"
```
volllly, line 32: **question:** why does `nbgv prepare-release` not work here?

.github/workflows/command-package.yml
```diff
@@ -70,7 +70,7 @@ jobs:
     
   deploy:
     if: (!always())
-    # if: (!failure()) && github.event.client_payload.slash_command.command == 'deploy'
+    #test if: (!failure()) && github.event.client_payload.slash_command.command == 'deploy'
```
volllly, line 73: **suggestion:** I think we can activate that now.
```suggestion
    if: (!failure()) && github.event.client_payload.slash_command.command == 'deploy'
```

storage/global.json
LeilaZarban: That's for testing. removed

storage/src/fiskaltrust.storage/version.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/main/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "1.3.21",
+  "version": "1.3.77",
```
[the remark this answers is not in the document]
LeilaZarban: The dsfinvk export uses 1.3.77

LeilaZarban (approved):

Issue #585, opened 2025-11-24
Improve country code handling in the queue


volllly:
## Background

In the v2 localization we currently don't check if the country code of the various Cases matches the country of the queue.
We also just take the country code of the receipt request and return that as the country code of the `ftState` so if the sent one is incorrect the received one is aswell.

## Feature

- Calculate the country code of the returned `ftState` from the queues country not from the `ftReceiptRequest`
- Fail with `0xEEEE_EEEE` `ftState` if the country code of the `ReceiptCase` does not match
- Fail with `0xEEEE_EEEE` `ftState` if the country code of the `ChargeItemCase` does not match
- Fail with `0xEEEE_EEEE` `ftState` if the country code of the `PayItemCase` does not match

## Tasks

- [ ] Implement validation for receipt request country code in localization v2
- [ ] Implement calculating the `ftState` country code from the queue country in localization v2
- [ ] Check how localization v1 (`fiskaltrust.Middleware.Queue`) handles wrong country codes in the receipt request and fix there as well if needed
- [ ] Implement calculating the `ftState` country code from the queue country in localization v1
- [ ] Add tests to v1 (`fiskaltrust.Middleware.Queue`) that test the behaviour
- [ ] Add tests to v2 localization that test the behaviour

PR #590, opened 2025-12-04
Improve country code handling in the queue #585 

pawelvds:

{Summary of the changes}

## Description

{Detail}

Fixes #{issue number}

### DIFF
--- a/queue/src/fiskaltrust.Middleware.Queue/SignProcessor.cs
+++ b/queue/src/fiskaltrust.Middleware.Queue/SignProcessor.cs
@@ -12,7 +12,6 @@ using fiskaltrust.storage.V0;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 
-
 namespace fiskaltrust.Middleware.Queue
 {
     public class SignProcessor : ISignProcessor
@@ -27,7 +26,6 @@ namespace fiskaltrust.Middleware.Queue
 
         private readonly MiddlewareConfiguration _middlewareConfiguration;
         private readonly SignatureFactory _signatureFactory;
-        //private readonly Action<string> _onMessage;
 
         public SignProcessor(
             ILogger<SignProcessor> logger,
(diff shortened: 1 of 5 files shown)


queue/src/fiskaltrust.Middleware.Localization.v2/SignProcessor.cs
```diff
@@ -193,6 +193,77 @@ private ReceiptResponse CreateReceiptResponse(ReceiptRequest receiptRequest, ftQ
...
+            {
+                foreach (var payItem in request.cbPayItems)
+                {
+                    var payItemCountry = payItem.ftPayItemCase.Country();
+                    if (!string.Equals(payItemCountry, queueCountry, StringComparison.OrdinalIgnoreCase))
+                    {
+                        return ReturnWithCountryMismatch("PayItemCase", payItemCountry, queue, receiptResponse, queueItem);
+                    }
+                }
+            }
```
volllly: **suggestion:** You could use the for loops above to perform both checks.

queue/src/fiskaltrust.Middleware.Queue/SignProcessor.cs
```diff
@@ -391,5 +452,18 @@ private async Task UpdateQueuesLastReceipt(ftQueue queue, ftReceiptJournal recei
...
+        {
+            return countryCode?.ToUpperInvariant() switch
+            {
+                "DE" => 0x4445000000000000,
+                "FR" => 0x4652000000000000,
+                "ME" => 0x4D45000000000000,
+                "IT" => 0x4954000000000000,
+                "AT" => 0x4154000000000000,
+                _    => 0
+            };
```
volllly: **thought:** I fear that when adding new countries we would forget to add it here.

Either throw an exception if it's an unknown country code.

Or you could use something like [we do here](https://github.com/fiskaltrust/middleware-interface-dotnet/blob/master/src/fiskaltrust.ifPOS/v2/Cases/ReceiptCase.cs#L84-L92) to calculate the String from the number.

volllly (changes requested):

PR #594, opened 2025-12-16
Misc localization v2 refactors

volllly:


queue/src/fiskaltrust.Middleware.Localization.QueueGR/Logic/ReceiptReferenceProvider.cs
```diff
@@ -6,6 +6,7 @@
 
 namespace fiskaltrust.Middleware.Localization.QueueGR.Logic;
 
+// This belongs in the storage or in the queuestorageprovider in the localization v2.
```
StefanKert: Yes, can you move it? thanks!

queue/src/fiskaltrust.Middleware.Localization.QueueBE/QueueBEBootstrapper.cs
```diff
@@ -23,24 +24,27 @@ public QueueBEBootstrapper(Guid id, ILoggerFactory loggerFactory, Dictionary<str
     {
         var middlewareConfiguration = MiddlewareConfigurationFactory.CreateMiddlewareConfiguration(id, configuration);
         var signaturCreationUnitBE = new ftSignaturCreationUnitBE();
+        // With the storage project in the middleware repo this _could_ already be done correctly.
```
StefanKert: Bit confused about this comment

queue/src/fiskaltrust.Middleware.Localization.v2/SignProcessor.cs
```diff
@@ -173,6 +173,7 @@ private ReceiptResponse CreateReceiptResponse(ReceiptRequest receiptRequest, ftQ
             cbReceiptReference = receiptRequest.cbReceiptReference,
             ftCashBoxIdentification = cashBoxIdentification,
             ftReceiptMoment = DateTime.UtcNow,
+            // Should we not take this from the `queue.CountryCode` and is there ever a valid case where this can be empty?
```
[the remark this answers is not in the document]
volllly: And do we really need the fallback? If its not set it _should_ fail, right?

queue/src/fiskaltrust.Middleware.Localization.QueueES/Processors/InvoiceCommandProcessorES.cs
```diff
@@ -86,8 +86,16 @@ public async Task<ProcessCommandResponse> FullInvoiceRequestAsync(ProcessCommand
...
+
         // Generate series identifier if not set
+        // Is there a convention for the fkt prefix?
+        // Why 1000?
+
+        // We'll have to check the supported length of the numseriesfactura in verifactu.
+
+        // That should not be initialized here but during the first initialization of the queue. Maybe in the bootstrapper or maybe in the start receipt.
+
+        // What is the cashboxidentification of the queue for if we don't use it here?
```
StefanKert: The CashBoxIdentification is the identifier of the so called device. The series is not necessarily the device which is why those things are seperated

queue/src/fiskaltrust.Middleware.Localization.QueueES/Processors/LifecycleCommandProcessorES.cs
```diff
@@ -7,18 +7,24 @@
...
     private readonly ILocalizedQueueStorageProvider _queueStorageProvider = queueStorageProvider;
-
+    private readonly AsyncLazy<IConfigurationRepository> _configurationRepository = configurationRepository;
     public async Task<ProcessCommandResponse> InitialOperationReceipt0x4001Async(ProcessCommandRequest request)
     {
         var (queue, receiptRequest, receiptResponse) = request;
         var actionJournal = ftActionJournalFactory.CreateInitialOperationActionJournal(receiptRequest, receiptResponse);
         await _queueStorageProvider.ActivateQueueAsync();
+        var queueES = await (await _configurationRepository).GetQueueESAsync(request.queue.ftQueueId);
+        queueES.InvoiceSeries = $"fkt{Helper.ShortGuid(request.queue.ftQueueId)}1000";
```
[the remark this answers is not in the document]
volllly: The initial operation receipt can't be executed twice so that can't be called twice 🤔

Or is this field meant to be already set during initialization through the `init_ftQueueES`? In that case should it not fail if it's not set?

queue/src/fiskaltrust.Middleware.Localization.v2/Storage/QueueStorageProvider.cs
```diff
@@ -62,8 +56,7 @@ public async Task<ftQueueItem> ReserveNextQueueItem(ReceiptRequest receiptReques
             cbReceiptReference = receiptRequest.cbReceiptReference,
             ftQueueRow = await IncrementQueueRow(),
             country = _cachedQueue.CountryCode,
-            // TOdo we need to set this to the correct procsesing version
-            // ProcessingVersion = _middlewareConfiguration.ProcessingVersion,
+            ProcessingVersion = ThisAssembly.AssemblyInformationalVersion,
```
StefanKert: Let's deploy this and see what version it writes. In some of the demos I ran it was somehow a weird mix between 1.0.0 and 1.3.0

volllly: In my tests it was the full semver version.

`ThisAssembly` is created by nerdbank and we did not have that before so I don't think that was what you had in the demos before 🤔

StefanKert (approved):

PR #643, opened 2026-03-24
#636 Add subtotal Discount / Surcharge

diego-santinelli-hrm:
- Improve for EpsonRT Printer "Sconto a subtotale"
- Fix negative import sent to Printer

- Added support for subtotal-level discounts and surcharges in the Epson RT printer fiscal receipt generation.             
- Fixed GetTotalAndMessages() to use Math.Abs() for negative payment amounts, ensuring correct handling when discounts cause negative pay item values  

#636 #638


accodev (changes requested):
The localflags proposed in this PR squash the localflags already in place for the RT devices.

https://docs.fiskaltrust.eu/docs/poscreators/middleware-doc/italy/reference-tables/ftreceiptcase

Please use 0x0100 and 0x0200 instead.

accodev (approved):

PR #654, opened 2026-04-03
Add noOp Logic and unit test

### LINKED ISSUE #581: Return successful ftState on receipts that are not signed
## Background

The "daily" receipts such as daily closing, monthly closing, yearly closing etc. should always return a successful response even if they are not actually signed/have no effect.

A POS should not need to know if a monthly closing is required or not in a market. It should in all markets just be able to send one and know that it's done what it needs to do.

## Problem

In some italian SCUs the SCU returns an Exception when it does not support this type of receipt. Instead it should just return a response with no signatures.

## Solution

If this is not related to the SCU and handled the same for all then we can just not send those kinds of receipt to the SCU.

If support for these receipts differs by SCU then we need to send them to the SCU and just return no signatures. 

## Open Questions

Is there a ftStateFlag that shows that this receipt had no effect?

### LINKED ISSUE #584: RT-Server doesn't accept RT-Printer specific flag
The RT Server implementation should be more resilient in case of sending cases that are not supported to keep compatibility with the overall implementation.

Currently the case `0x4954_2003_0000_3000` is not supported and therefore leads to an error. In cases of unsupported we should return a NoOp.

diego-santinelli-hrm:
Fixes #581 and #584 in `CustomRTServerSCU`: both issues had the same root cause: unhandled receipt cases were throwing an exception instead of returning gracefully.

- ProtocolUnspecified0x3000 → NoOp 
 - MonthlyClosing / YearlyClosing  routed to `PerformDailyCosingAsync`, same as `DailyClosing`
 - Final throw replaced with NoOp for unknown/future cases

Fixes #584 #581

### DIFF
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.CustomRTServer/CustomRTServerSCU.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.CustomRTServer/CustomRTServerSCU.cs
@@ -109,11 +109,16 @@ public sealed class CustomRTServerSCU : LegacySCU
                 return ProcessResponseHelpers.CreateResponse(request.ReceiptResponse, signatures);
             }
 
-            if (request.ReceiptRequest.IsDailyClosing())
+            if (request.ReceiptRequest.IsDailyClosing() || request.ReceiptRequest.IsMonthlyClosing() || request.ReceiptRequest.IsYearlyClosing())
             {
                 (var signatures, var state) = await PerformDailyCosingAsync(request.ReceiptRequest, request.ReceiptResponse, cashuuid);
                 request.ReceiptResponse.ftState = state;
                 return ProcessResponseHelpers.CreateResponse(request.ReceiptResponse, signatures);
+            }
+
+            if (receiptCase == (long) ITReceiptCases.ProtocolUnspecified0x3000)
+            {
+                return ProcessResponseHelpers.CreateResponse(request.ReceiptResponse, new List<SignaturItem>());
             }
 
             switch (receiptCase)
@@ -127,7 +132,7 @@ public sealed class CustomRTServerSCU : LegacySCU
                     return ProcessResponseHelpers.CreateResponse(request.ReceiptResponse, signatures);
             }
 
-            throw new Exception($"The given receiptcase 0x{receiptCase.ToString("X")} is not supported.");
+            return ProcessResponseHelpers.CreateResponse(request.ReceiptResponse, new List<SignaturItem>());
         }
         catch (Exception ex)
         {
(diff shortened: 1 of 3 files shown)


scu-it/src/fiskaltrust.Middleware.SCU.IT.CustomRTServer/CustomRTServerSCU.cs
```diff
@@ -127,7 +132,7 @@ public override async Task<ProcessResponse> ProcessReceiptAsync(ProcessRequest r
                     return ProcessResponseHelpers.CreateResponse(request.ReceiptResponse, signatures);
             }
 
-            throw new Exception($"The given receiptcase 0x{receiptCase.ToString("X")} is not supported.");
+            return ProcessResponseHelpers.CreateResponse(request.ReceiptResponse, new List<SignaturItem>());
         }
```
accodev, line 136: can you briefly explain what is the purpose of this PR?

accodev (dismissed):

scu-it/src/fiskaltrust.Middleware.SCU.IT.CustomRTServer/CustomRTServerSCU.cs
```diff
@@ -109,13 +109,18 @@ public override async Task<ProcessResponse> ProcessReceiptAsync(ProcessRequest r
                 return ProcessResponseHelpers.CreateResponse(request.ReceiptResponse, signatures);
             }
 
-            if (request.ReceiptRequest.IsDailyClosing())
+            if (request.ReceiptRequest.IsDailyClosing() || request.ReceiptRequest.IsMonthlyClosing() || request.ReceiptRequest.IsYearlyClosing())
```
StefanKert, line 112: How are we handling this in the Epson Printer? I am not sure if we should expect MonthlyClosing & YearlyClosing to also do a DailyClosing. Usually these operations are cauled in sequence:

31. December
1. Daily Closing (31 of dec)
2. Monthly Closing (closing december)
3. Yearly Closing (closing the year)

So we would do three daily closings in this scenario

diego-santinelli-hrm: Screenshot 2026-04-08 095641
 In the Epson SCU we handle it the same way — all three (Daily, Monthly, Yearly)
  call PerformDailyCosing() as separate if blocks (see EpsonRTPrinterSCU.cs lines
  96-109). So yes, in the Dec 31 scenario we'd get three daily closings, which is
  the expected behavior since each closing triggers a Z report on the fiscal device

StefanKert (approved):

PR #666, opened 2026-05-16
Fix QueueGR myDATA aa numbering skips for non-submitted receipts

StefanKert:
## Summary

Fixes [fiskaltrust/market-gr#172](https://github.com/fiskaltrust/market-gr/issues/172): QueueGR was deriving the AADE `invoiceHeader.aa` from the generic `ftReceiptNumerator`, which advances for every successful receipt regardless of whether it was actually submitted to myDATA. QueueGR routes daily/monthly/yearly closings, zero receipts, protocol events, and other receipt cases through `GRFallBackOperations.NoOp` — those advance `ftReceiptNumerator` but never reach AADE, so the next real submission sees `aa = N+2` instead of `N+1` and the merchant's fiscal sequence has gaps.

Phase 1 fix:

- **Dedicated `InvoiceSeries` / `InvoiceNumerator` on `ftQueueGR`** (1:1 shape match with `ftQueueES`). Only ever advances on actual AADE submission.
- **Reserve-then-commit-on-`State.Success`** in `ReceiptCommandProcessorGR` / `InvoiceCommandProcessorGR`, copying the ES pattern. Skips the commit when the caller bypasses our counter via handwritten payload or `mydataoverride` (detected by suffix mismatch on `ftReceiptIdentification`).
- **AADE error 233 → `State.Success`** in `MyDataSCU`. Without this, a process kill between AADE's 200 OK and our storage commit leaves the queue stuck forever; with it, the next retry advances the counter and the queue self-heals.
- **One-time migration on first post-upgrade submission**: seeds `InvoiceNumerator` from `ftReceiptNumerator` so the first new `aa` lands strictly above anything AADE already has on file for the queue.
- **Activation seeding** of `InvoiceSeries` from `CashBoxIdentification` for fresh queues (parity with `LifecycleCommandProcessorES.cs:24-30`).

`ftReceiptNumerator` semantics are unchanged for every country, including GR.

## Scope notes

- Storage model is touched only on Azure Table Storage and InMemory. SQLite / MySQL / EF `QueueGR` repositories already `throw new NotImplementedException()` for GR and are intentionally left alone.
- `MyDataSCU`'s sandbox `GovernmentApi` stash on `ftStateData` is updated to merge rather than replace, so the country-processor's `ProposedInvoiceCounter` survives the SCU call.
- `AADEFactory` keeps the legacy `ftReceiptIdentification`-parse path as a fallback so existing tests and direct callers continue to work without modification.
- The QueueGR project now project-references `scu-gr/MyData` for the shared `MiddlewareSCUGRMyDataState` POCO. Namespace sharing in that direction already existed implicitly (`fiskaltrust.Middleware.Localization.QueueGR.Validation` is declared by files in scu-gr); this makes the dependency explicit.

## Out of scope (Phase 2, tracked separately)

- Mark recovery via `/myDataProvider/RequestTransmittedDocs` on 233 (to repopulate the prior submission's `mark` / `uid` / `authenticationCode` on the resubmitted response).
- Content-fingerprint verification to distinguish crash-and-retry (case A) from foreign collision (case B) and surface case B via `ftActionJournal`.

Phase-1 advance-on-233 behaviour is correct in practice (`series` is queue-scoped, so foreign collisions imply a deployment misconfiguration). Phase 2 is operational-quality work on top, not a correctness requirement.

## Test plan

- [x] `dotnet test scu-gr/test/.../MyData.UnitTest` — 804 tests pass (no regressions).
- [x] `dotnet test queue/test/.../QueueGR.UnitTest` — 42 tests pass (34 pre-existing + 8 new).
- [x] New `InvoiceCounterReservationTests` covers: success advances the counter, error doesn't, handwritten/override suffix mismatch doesn't, success without mark signature still commits with `LastInvoiceMark = null`, empty `InvoiceSeries` falls back to `CashBoxIdentification`, upgrade migration seeds from `ftReceiptNumerator`.
- [x] New `LifecycleCommandProcessorGRTests` cases for `InvoiceSeries` seeding (fresh queue + idempotency).
- [ ] Manual smoke against the AADE sandbox once the branch is built — verify `POS → DailyClosing → POS` produces `aa = N, N+1` on the AADE side.


queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
```diff
@@ -0,0 +1,94 @@
...
+        // produced (we overshoot by the count of NoOp receipts that previously
+        // advanced ftReceiptNumerator without submitting) is acceptable — AADE
+        // permits non-contiguous aa, it just refuses duplicates.
+        if (string.IsNullOrEmpty(queueGR.InvoiceSeries))
+        {
+            queueGR.InvoiceSeries = queueGR.CashBoxIdentification;
+        }
+        if (queueGR.InvoiceNumerator == 0 && request.queue.ftReceiptNumerator > 0)
+        {
+            queueGR.InvoiceNumerator = request.queue.ftReceiptNumerator;
```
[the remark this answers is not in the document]
StefanKert: Fixed in 11b69a63. Migration now gates on `string.IsNullOrEmpty(queueGR.InvoiceSeries)` rather than on the counter value — `LifecycleCommandProcessorGR` seeds `InvoiceSeries` at activation for new queues, so any queue reaching the reservation helper with `InvoiceSeries` still empty is definitionally pre-upgrade. Brand-new queues with prior NoOps now correctly start at `aa = 1`. Added `NewQueueWithNoOpsBeforeFirstSubmission_StartsAtAa1` as a regression test.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/MyDataSCU.cs
```diff
@@ -360,6 +360,20 @@ private void SetErrorAndLog(ProcessRequest request, string errorMessage)
         request.ReceiptResponse.SetReceiptResponseError(errorMessage);
     }
 
+    private static void SetCountrySuffix(ReceiptResponse response, string series, string aa)
```
StefanKert, line 370: This is a bit confusing.. If we use handwritten series + number the queue should already take this inbound so this is something that we shoul dnever see. In cases where these things are off we might have an actual problem

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
```diff
@@ -0,0 +1,212 @@
...
+        // The seed is persisted immediately so it survives a failed first submission and
+        // the history scan runs at most once per queue.
+        //
+        // Deliberate residual risk: an attempt that AADE accepted but that was stored as
+        // failed on our side (timeout/crash between AADE's 200 OK and persisting the
+        // response) has no success response in the history, so its aa is re-reserved and
+        // AADE will reject the resubmission with error 233 (duplicate uid) instead of us
+        // skipping a number. Gap-free continuity is the requirement here; recovering
+        // from 233 is the planned self-heal follow-up.
+        if (string.IsNullOrEmpty(queueGR.InvoiceSeries))
```
StefanKert, line 46: Can we add this migration to the start of the queue? so that we can make sure that it blows up if we can't migrat it and it stil uses the old waay

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/MyDataSCU.cs
StefanKert: Addressed in d93ae8f3: the queue now takes the handwritten (Series, AA) inbound from ftReceiptCaseData and stamps them into ftReceiptIdentification before the SCU is called — no reservation and no counter access for handwritten receipts, so this rewrite can no longer change anything for them. SetCountrySuffix itself stays: Order0x3004/Pay0x3005 call the SCU without a pre-appended segment (it preserves their old append behaviour), and it keeps the invariant that a success response's segment always reflects what actually went to AADE. The only remaining path with a differing segment is a mydataoverride carrying its own Series/Aa — caller-numbered as well, so the counter commit is skipped for those.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
StefanKert: Done in d93ae8f3: the migration moved to InvoiceCounterMigration and runs at queue start — started eagerly at queue construction and awaited before every receipt. If it cannot run, receipt processing fails loudly (and the reservation additionally throws if it ever sees an uninitialized InvoiceSeries), so the queue can never silently fall back to the legacy ftReceiptNumerator-derived numbering.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Models/Cases/SignatureTypeGR.cs
```diff
@@ -7,6 +7,7 @@ public enum SignatureTypeGR : long
 {
     InitialOperationReceipt = 0x4752_2000_0000_0003,
     OutOfOperationReceipt = 0x4752_2000_0000_0003,
+    Mark = 0x4752_2000_0000_0014,
```
StefanKert, line 10: Does this correlate with what we have in the SCUGR?

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/MyDataSCU.cs
```diff
@@ -297,7 +297,6 @@ public async Task<ProcessResponse> ProcessReceiptAsync(ProcessRequest request, L
                         var enrichedPayload = AADEFactory.GenerateInvoicePayload(doc);
                         // Use the downloadingInvoiceUrl from the invoice for the QR code
                         request.ReceiptResponse.AddSignatureItem(SignatureItemFactoryGR.CreateGRQRCode(doc.invoice[0].downloadingInvoiceUrl));
-                        request.ReceiptResponse.ftReceiptIdentification += $"{doc.invoice[0].invoiceHeader.series}-{doc.invoice[0].invoiceHeader.aa}";
```
StefanKert, line 300: Is it safe to remove this?

queue/src/fiskaltrust.Middleware.Localization.QueueGR/QueueGRBootstrapper.cs
```diff
@@ -29,8 +29,25 @@ public QueueGRBootstrapper(Guid id, ILoggerFactory loggerFactory, Dictionary<str
...
-        var signProcessor = new SignProcessor(loggerFactory.CreateLogger<SignProcessor>(), queueStorageProvider, signProcessorGR.ProcessAsync, cashBoxIdentification, middlewareConfiguration);
+        var configurationRepository = storageProvider.CreateConfigurationRepository();
+        var queueItemRepository = storageProvider.CreateMiddlewareQueueItemRepository();
+
+        // One-time invoice-counter migration, started eagerly at queue construction
+        // (AsyncLazy runs its factory immediately) and awaited before every receipt. If
+        // the counter cannot be initialized, receipt processing fails loudly instead of
+        // silently falling back to the legacy ftReceiptNumerator-derived numbering.
+        var invoiceCounterMigration = new AsyncLazy<bool>(async () =>
```
StefanKert: **[major — flagged independently by 3 reviewers]** A faulted migration task is cached forever — one transient storage error at startup fails every receipt until process restart.

`AsyncLazy<T>` wraps a `Lazy<Task<T>>` around a `Task.Run` started at construction and never re-executes the factory. `EnsureMigratedAsync` performs many storage round-trips (config reads plus the per-row history walk; only `JsonException` is swallowed inside the loop). A single transient Azure Table timeout/throttle during the scan faults the cached task, and every subsequent receipt's `await invoiceCounterMigration` rethrows the same stale exception — `SignProcessor` persists 100% of traffic as failed until someone restarts the host, even though storage recovered seconds later. Before this PR a transient blip only failed the receipts issued during the blip.

Failing loudly is right; failing *permanently* isn't. Suggest re-creating the lazy on fault (retry on next receipt) or retry-with-backoff inside `EnsureMigratedAsync`.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterMigration.cs
```diff
@@ -0,0 +1,96 @@
...
+        // receipts) and submissions in a foreign series (handwritten / mydataoverride)
+        // are skipped. Error responses are skipped too: failed attempts must never count
+        // as submitted.
+        //
+        // The scan is unbounded on purpose. Capping it and falling back to a guess
+        // could re-issue an aa that AADE already has on file. The cost self-limits:
+        // queues that submit regularly find the segment within the last few rows, and
+        // queues that never submitted are small because they only ever collected NoOps.
+        // The scan runs once per queue — the seed is persisted right after.
+        for (var row = lastQueueRow; row >= 1; row--)
```
StefanKert, line 83: **[major — flagged independently by 3 reviewers]** The unbounded backward walk multiplies with `GetByQueueRowAsync` being a full-table scan on Azure Table Storage.

`AzureTableStorageQueueItemRepository.GetByQueueRowAsync` filters on `ftQueueRow`, a plain property — while the QueueItem PartitionKey is `Mapper.GetHashString(ftQueueRow)`, so each call scans the table server-side instead of doing a partition read. The loop makes one such call per row walked, i.e. **O(rowsWalked × tableSize)**. The "cost self-limits" rationale has two holes: (a) a queue whose recent submissions are all foreign-series (handwritten-heavy merchants) walks past every one of them; (b) a queue with years of closings/zero receipts and no own-series success walks all rows. During the whole scan every receipt blocks on `await invoiceCounterMigration`, and a restart begins from scratch since the seed is only persisted at the end.

Since the partition key is derivable from the row number, querying by `PartitionKey == Mapper.GetHashString(row)` (or a paged descending query) turns each lookup into a cheap partition-scoped read.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterMigration.cs
```diff
@@ -0,0 +1,96 @@
...
+                continue;
+            }
+            if (response == null || !response.ftState.IsState(State.Success))
+            {
+                continue;
+            }
+            if (CountrySegment.TryParse(response.ftReceiptIdentification, out var series, out var aa)
+                && string.Equals(series, invoiceSeries, StringComparison.Ordinal))
+            {
+                return aa;
```
StefanKert: **[major]** The seed can be poisoned by historical aa-only `mydataoverride` submissions — and the resulting 233s wedge the queue permanently.

The seed takes the *newest* own-series success segment at face value. But the pre-PR `MyDataSCU` appended the **submitted** values after `ApplyInvoiceHeaderOverride` ran, and the old override logic applied `series` and `aa` independently — a historical override of only `aa` (series stayed = `ftCashBoxIdentification`) produced an own-series segment with an arbitrary caller-chosen aa. If that receipt is the newest own-series success:
- override aa *above* the real sequence → permanent forward jump/gap;
- override aa *below* it → next reservation collides, AADE answers 233, the error response has no mark, the counter never advances — **every subsequent submission retries the same aa forever** (233 recovery is explicitly deferred).

The same no-mark-on-233 mechanic applies to the documented crash-window residual risk above: it doesn't cost one receipt, it wedges the queue until the self-heal ships. Consider seeding from the **max** own-series aa rather than the newest, and logging loudly when the seed diverges from the `ftReceiptNumerator`-derived expectation.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
```diff
@@ -0,0 +1,144 @@
...
+    {
+        series = string.Empty;
+        aa = 0;
+        if (!request.ftReceiptCase.IsFlag(ReceiptCaseFlags.HandWritten))
+        {
+            return false;
+        }
+        if (!request.TryDeserializeftReceiptCaseData<ftReceiptCaseDataPayload>(out var payload)
+            || string.IsNullOrEmpty(payload?.GR?.Series)
+            || payload!.GR!.AA is not > 0)
```
StefanKert, line 93: **[major — flagged independently by 3 reviewers]** The handwritten AA gate diverges from the SCU's validation: the queue requires `AA > 0`, but `AADEFactory` only rejects `null`/`0` (`data?.GR?.AA == null || data?.GR?.AA == 0`, AADEFactory.cs:275) — a **negative** AA passes there.

Concrete failure: HandWritten receipt with `GR.Series="HW", AA=-5` and a consistent `HashPayload` (the hash covers the caller's own `-5`, so it validates). The queue rejects the payload as incomplete → falls through to the reservation path and appends `#CB-A-42`; the SCU accepts the payload and files `HW`/`-5` at AADE (`aa` is a plain string in the myDATA schema). AADE answers Success + invoiceMark → the commit gate advances `InvoiceNumerator` to 42 — an aa AADE **never received** — recreating exactly the numbering skip this PR eliminates, with `ft…#CB-A-42` persisted as the identification of a document actually filed as `HW`/`-5`.

Align the two gates (reject `AA <= 0` in the SCU's handwritten validation, or accept the SCU's predicate here) so a payload can never be "incomplete" for the queue but "valid" for the SCU.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
```diff
@@ -0,0 +1,144 @@
...
+    public static async Task<ProcessCommandResponse> InvokeWithCounterAsync(
+        ProcessCommandRequest request,
+        AsyncLazy<IConfigurationRepository> configurationRepository,
+        Func<Task<ProcessResponse>> sscdCall)
+    {
+        // Handwritten documents are caller-numbered: the merchant already stamped
+        // (series, aa) on the paper original, so the queue takes them inbound verbatim.
+        // No reservation is made, the counter never advances and no storage is touched.
+        // Full validation of the handwritten payload stays in the SCU.
+        if (TryGetHandwrittenNumbering(request.ReceiptRequest, out var handwrittenSeries, out var handwrittenAa))
```
StefanKert, line 30: **[minor]** Nothing rejects a handwritten payload whose `Series` equals the queue's own `InvoiceSeries` (= `CashBoxIdentification`). If a caller submits such a receipt with an AA at/above the current numerator and AADE accepts it, the counter is unaware: a later automatic reservation reserves the same `(series, aa)` → AADE 233 → no mark → the counter never advances past it (same wedge mechanic as the migration-seed case).

The PR closes exactly this hole for `mydataoverride` (series/aa overrides now throw in `ApplyInvoiceHeaderOverride`) but leaves it open for handwritten numbering. Cheap guard: reject handwritten payloads whose `Series` ordinal-equals `queueGR.InvoiceSeries`.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
```diff
@@ -0,0 +1,144 @@
...
+        // (handwritten numbering is taken inbound, series/aa overrides are rejected)
+        // and AADEFactory derives the document numbering from that segment.
+        var mark = TryExtractMark(response.ReceiptResponse);
+        if (response.ReceiptResponse.ftState.IsState(State.Success) && mark != null)
+        {
+            queueGR.InvoiceNumerator = reservedAa;
+            queueGR.LastInvoiceMoment = request.ReceiptRequest.cbReceiptMoment;
+            queueGR.LastInvoiceQueueItemId = response.ReceiptResponse.ftQueueItemID;
+            queueGR.LastInvoiceMark = mark;
+            await configRepo.InsertOrUpdateQueueGRAsync(queueGR);
```
StefanKert, line 77: **[minor]** If `InsertOrUpdateQueueGRAsync` throws after the SCU returned Success + mark, the invoice exists at AADE under the reserved `(series, aa)` but the counter never advanced: the receipt is persisted as failed (segment kept), and every retry re-reserves the same aa — which AADE rejects as 233 — until the counter is fixed manually. The window is narrow and `ReceiptCommandProcessorES` shares the structure, so this is an accepted-risk class rather than a regression — but note the planned 233 self-heal must treat *this* direction as "number consumed, advance" (not "re-usable"). Logging mark + reserved aa before the write would let operators reconcile.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
```diff
@@ -0,0 +1,144 @@
...
+    {
+        // AADE's invoiceMark as stamped by the SendInvoices success path, which types it
+        // as SignatureTypeGR.Mark. The SCU's non-invoice flows (delivery-note
+        // cancellation, payment methods) type all their response items as
+        // GenericMyDataInfo, so they can never satisfy this lookup — even if one of
+        // their items happens to be captioned "invoiceMark".
+        var markSignature = response.ftSignatures?
+            .FirstOrDefault(s => string.Equals(s.Caption, "invoiceMark", StringComparison.Ordinal)
+                && s.ftSignatureType.IsType(SignatureTypeGR.Mark));
+        return markSignature != null && long.TryParse(markSignature.Data, out var mark)
```
StefanKert: **[minor]** `long.TryParse(markSignature.Data, out var mark)` uses `NumberStyles.Integer` with the current culture, while `CountrySegment.TryParse` in this same PR deliberately uses `NumberStyles.None, CultureInfo.InvariantCulture`. The default overload accepts inputs a mark can never legitimately be (`" 123"`, `"-5"`) and can behave differently under exotic host cultures. Use the same strict invariant parse here.

StefanKert (commented):
Multi-agent review (5 reviewers: reservation correctness, SCU-side, storage/migration, test coverage, cross-cutting). Findings were cross-verified against the checked-out head (446ad03) before posting; three of the majors were flagged independently by 2–3 reviewers.

**PR description is out of sync with the diff.** The body claims things this change does not contain:
- No AADE error **233 → `State.Success`** mapping exists in `MyDataSCU` (the migration comment explicitly defers 233 recovery). Several inline findings below hinge on this: a 233 never advances the counter, so every "AADE has it, we don't" divergence *wedges* the queue on one aa rather than losing a single receipt.
- No **InMemory** storage changes and no **`ftStateData` merge** are in the diff.
- No **QueueGR→scu-gr project reference** — the csproj only gained `InternalsVisibleTo`; the shared contract (segment convention, `Mark` enum value, `"invoiceMark"` caption) is duplicated by convention.
- The migration seeds from the **queue-item history scan**, not from `ftReceiptNumerator` as the body states.

**Top themes** (details inline):
1. **Startup migration robustness**: a faulted `AsyncLazy` is cached until process restart, and the history walk does per-row Azure Table *filter* queries (not point reads) with no bound — both convert a transient blip or a NoOp-heavy history into a fully unavailable queue.
2. **Two gates that must agree but don't**: the queue accepts handwritten `AA > 0`, the SCU accepts `AA != 0`; and queue/SCU are separately shipped packages with no version gate on the new segment contract.
3. **233-wedge class**: migration seed poisoning, handwritten collisions with the own series, and commit-write failures all funnel into "AADE consumed the aa, the counter didn't" — currently unrecoverable without manual intervention.

Test coverage of the new logic is otherwise solid (recovery paths, continuity, retry-reuse are all pinned).

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Models/Cases/SignatureTypeGR.cs
StefanKert, line 10: **[minor]** The commit gate rests on this value and the `"invoiceMark"` caption matching what scu-gr produces — but `Mark = 0x…0014` duplicates scu-gr's own `SignatureTypeGR` enum, and the caption comes from `ItemsChoiceType.invoiceMark.ToString()`; there is no shared constant and no cross-project test pinning either side.

If they ever drift (enum renumbering, schema regeneration, caption change), `TryExtractMark` returns null on every Success, the counter **silently stops advancing**, and every subsequent invoice submits the same reserved aa → production 233 rejections with no log hint. Values match today (verified both enums and the caption source) — but consider a shared constant or a pinning test, plus a warning log when a Success on an invoice-filing path carries no recognizable mark.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/MyDataSCU.cs
```diff
@@ -297,7 +297,6 @@ public async Task<ProcessResponse> ProcessReceiptAsync(ProcessRequest request, L
                         var enrichedPayload = AADEFactory.GenerateInvoicePayload(doc);
                         // Use the downloadingInvoiceUrl from the invoice for the QR code
                         request.ReceiptResponse.AddSignatureItem(SignatureItemFactoryGR.CreateGRQRCode(doc.invoice[0].downloadingInvoiceUrl));
```
StefanKert, line 312: **[major]** Queue and SCU are separately deployed packages — a version skew silently breaks the new segment contract in both directions.

- **New queue + old SCU**: the queue appends `ft1#CB-A-42`, the old `AADEFactory` still parses the hex prefix (files `series=cashbox, aa=1`), and the old SCU emits the same `invoiceMark`/`SignatureTypeGR.Mark` signature that satisfies the new commit gate — so the queue commits `InvoiceNumerator=42` while AADE received `aa=1`, every receipt. The old SCU also appends its own segment, corrupting the identification (`ft1#CB-A-42CB-A-1`).
- **New SCU + old queue**: successful submissions no longer get the segment persisted, so a later queue upgrade seeds `InvoiceNumerator` from a pre-skew row and re-reserves aa values AADE already has → every receipt fails with 233 (self-heal not implemented).

This needs at least a hard compatibility note in the release (min-version pairing), and ideally a loud failure when the SCU sees a receipt case that requires the segment but finds none, rather than the silent legacy fallback.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEFactory.cs
```diff
@@ -627,16 +627,16 @@ private static void ApplyInvoiceHeaderOverride(AadeBookInvoiceType invoice, Invo
...
+        if (!string.IsNullOrEmpty(headerOverride.Series) || !string.IsNullOrEmpty(headerOverride.Aa))
         {
-            invoice.invoiceHeader.series = headerOverride.Series;
-        }
-
-        // Apply aa (sequential number)
-        if (!string.IsNullOrEmpty(headerOverride.Aa))
-        {
-            invoice.invoiceHeader.aa = headerOverride.Aa;
+            throw new ArgumentException(
```
StefanKert, line 642: **[minor]** This is a breaking change for existing `mydataoverride` integrations: `invoiceHeader.series`/`aa` were previously honored (documented model members, covered by the old `MapToInvoicesDoc_WithHeaderIdentificationOverrides_ShouldSetFields` test) — now any request carrying either field errors on every receipt. The rejection is the right call for counter integrity, but it needs an explicit release note/migration hint: the error's pointer to the HandWritten flag only covers the paper-document use case, not integrations that used the override to control numbering.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEFactory.cs
```diff
@@ -1725,4 +1725,35 @@ public static string GeneratePaymentMethodPayload(PaymentMethodsDoc doc)
...
+
+    private static (string series, string aa) ResolveSeriesAndAa(ReceiptResponse receiptResponse)
+    {
+        // Preferred path: the QueueGR processor pre-appends "{series}-{aa}" to
+        // ftReceiptIdentification before invoking the SCU. This mirrors the convention
+        // every other country queue uses (ES/FR/AT/PT all append a country segment
+        // after the "#"). The handwritten and mydataoverride paths overwrite
+        // invoiceHeader.series / .aa on the doc afterwards via the existing override
+        // logic; the suffix on ftReceiptIdentification stays as the reserved values
+        // and is rewritten by MyDataSCU after AADE confirms the actually-submitted
```
StefanKert: **[minor]** Stale comment — both claims here describe mechanisms this same PR removed:
- the mydataoverride path no longer overwrites series/aa; `ApplyInvoiceHeaderOverride` now **throws** for them;
- the "rewritten by MyDataSCU after AADE confirms" step is the exact line deleted from `MyDataSCU.cs`, and nothing rewrites or compares the suffix anywhere.

This matters beyond style: it implies a rewrite-and-compare safeguard exists that would catch a header/segment divergence (see the handwritten-AA gate mismatch), but none does. The test comment in `AADEFactoryTests.MapToInvoicesDoc_Handwritten_OverridesSuffixBasedSeriesAndAa` repeats the same nonexistent "suffix-rewrite-and-compare safeguard". Please update both to the actual contract: the queue is the single writer of the segment and skips reservation entirely for handwritten documents.

queue/src/fiskaltrust.Middleware.Storage.AzureTableStorage/Repositories/Configuration/AzureTableStorageQueueGRRepository.cs
```diff
@@ -47,6 +47,11 @@ protected override AzureTableStorageFtQueueGR MapToAzureEntity(ftQueueGR src)
                 UsedFailedMomentMin = src.UsedFailedMomentMin?.ToUniversalTime(),
                 UsedFailedMomentMax = src.UsedFailedMomentMax?.ToUniversalTime(),
                 UsedFailedQueueItemId = src.UsedFailedQueueItemId,
+                InvoiceSeries = src.InvoiceSeries,
+                InvoiceNumerator = src.InvoiceNumerator,
+                LastInvoiceMoment = src.LastInvoiceMoment?.ToUniversalTime(),
```
StefanKert, line 52: **[minor]** `LastInvoiceMoment` is set from `cbReceiptMoment` (`InvoiceCounterReservation.cs:57`), which is deserialized from caller JSON — an ISO timestamp without offset yields `DateTimeKind.Unspecified`, and `?.ToUniversalTime()` then treats it as server-local and shifts it by the host's UTC offset before persisting. The neighboring fields mapped this way (`UsedFailedMomentMin` etc.) are `DateTime.UtcNow`-sourced, so the pattern was safe for them but isn't here. Informational field only, but if it's ever used to diagnose numbering incidents the skew will mislead — consider persisting the queue-side `ftReceiptMoment` (always UTC) instead.

queue/test/fiskaltrust.Middleware.Localization.QueueGR.UnitTest/Processors/InvoiceCounterReservationTests.cs
```diff
@@ -0,0 +1,607 @@
...
+        var processor = new ReceiptCommandProcessorGR(
+            grSSCDMock.Object,
+            Mock.Of<IQueueStorageProvider>(),
+            new AsyncLazy<IConfigurationRepository>(() => Task.FromResult(configRepoMock.Object)));
+
+        var request = BuildRequest(queue, ReceiptCase.PointOfSaleReceipt0x0001);
+
+        var result = await processor.PointOfSaleReceipt0x0001Async(request);
+
+        result.receiptResponse.ftState.IsState(State.Success).Should().BeTrue();
```
StefanKert, line 44: **[test gap]** The success path never asserts the final `ftReceiptIdentification` keeps the reserved segment. Every identification assertion in this file checks the *restore* path (`"ft1#"`) or the handwritten path; this test only checks `ftState` and the repo write, and the Order0x3004 test captures the identification *inside* the SCU mock, pre-restore-logic.

If `SubmitWithSegmentAsync` regressed to restore `originalReceiptIdentification` unconditionally, all tests would still pass — but the POS would stop receiving the series/aa, and `InvoiceCounterMigration.GetLastSubmittedAaAsync` (which seeds from exactly this persisted segment) would find nothing and re-seed at 0, re-issuing aa values AADE already has. Add `result.receiptResponse.ftReceiptIdentification.Should().Be("ft1#CB-A-42")` here.

queue/test/fiskaltrust.Middleware.Localization.QueueGR.UnitTest/Processors/InvoiceCounterReservationTests.cs
```diff
@@ -0,0 +1,607 @@
...
+        {
+            await processor.PointOfSaleReceipt0x0001Async(BuildRequest(queue, ReceiptCase.PointOfSaleReceipt0x0001));
+        }
+
+        capturedAa.Should().Equal(1L, 2L, 3L, 4L, 5L);
+        queueGR.InvoiceNumerator.Should().Be(5);
+    }
+
+    [Fact]
+    public async Task RetryAfterErrorReusesSameAa()
```
StefanKert, line 523: **[test gap]** The failure mode the reserve-then-commit design *creates* is untested: a crash/timeout between AADE's 200 OK and persisting the response stores the receipt as failed, this test proves the same aa is re-reserved — and AADE will then reject every subsequent submission with 233 (duplicate) forever, since the counter only advances on Success+mark and no 233 handling exists in this diff. At minimum add a test pinning the current (stuck) behavior so the Phase-2 self-heal has a red/green target.

scu-gr/test/fiskaltrust.Middleware.SCU.GR.MyData.UnitTest/AADEFactoryTests.cs
```diff
@@ -283,6 +283,164 @@ public void MapToInvoicesDoc_Should_Not_ThrowException_IfHandwritten_AndMerchant
         error!.Exception.Message.Should().Be("When using Handwritten receipts the MerchantVATID that is provided must match with the one configured in the Account.");
     }
 
+    [Fact]
+    public void MapToInvoicesDoc_WithCountrySuffix_ParsesSeriesAndAaFromSuffix()
```
StefanKert, line 287: **[test gap]** The single-writer invariant on `ftReceiptIdentification` is untested at the SCU level. The PR removes the append in `MyDataSCU.ProcessReceiptAsync`, and the whole queue-side commit/migration logic now assumes the queue is the only writer of the segment — but no test exercises `MyDataSCU`'s SendInvoices success path (the only `MyDataSCU` instantiation in tests fails validation before any HTTP call), and queue-side tests mock `IGRSSCD` entirely.

If the removed append were reintroduced (it existed for years — a bad merge could bring it back), all 800+ SCU tests and all 42 queue tests still pass, but production identifications become `ft1#CB-A-42CB-A-42`, and `CountrySegment.TryParse` would parse series `CB-A-42CB-A` on the next migration, corrupting the seed. A test driving `ProcessReceiptAsync` with a stubbed HTTP handler, asserting `ftReceiptIdentification` is left untouched, would pin this.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/MyDataSCU.cs
StefanKert: Currently only deployed as a single unit

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Models/Cases/SignatureTypeGR.cs
StefanKert: Yes — scu-gr’s SignatureTypeGR.Mark is 0x4752_2000_0000_0014, identical to this value (the queue duplicates the enum like the other GR case types). Since c32b3d1b both sides pin the contract with tests (MarkSignatureContract_*), so a drift fails the build instead of silently stopping the counter.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/MyDataSCU.cs
StefanKert: Yes — deliberately removed in 446ad039: the queue is the single writer of the segment (reserved values for automatic receipts, handwritten taken inbound, series/aa overrides rejected), and AADEFactory derives the document numbering from the segment itself — this rewrite could only ever write back what was already there. Whether an invoice was actually filed is decided by the invoiceMark gate instead. c32b3d1b adds a MyDataSCU-level test (stubbed HTTP) pinning that ProcessReceiptAsync leaves ftReceiptIdentification untouched on success.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/QueueGRBootstrapper.cs
StefanKert: Fixed in c32b3d1b: InvoiceCounterMigrationGate replaces the AsyncLazy — still started eagerly at construction and awaited before every receipt, but a faulted attempt is retried on the next receipt, so a transient storage error at startup only fails the receipts processed during the outage. Tests: Gate_RetriesFaultedMigration_OnNextReceipt, Gate_RunsMigrationOnlyOnce_WhenItSucceeds.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterMigration.cs
StefanKert: Fixed in c32b3d1b: GetByQueueRowAsync now queries by the ftQueueRow-derived PartitionKey (Mapper.GetHashString), turning each lookup into a partition-scoped read — the walk is O(rows) point reads, one-time per queue, and the gate retries it if interrupted.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterMigration.cs
StefanKert: Fixed in c32b3d1b: the seed is now the MAXIMUM own-series aa across the whole history, not the newest — collision-free by construction in both directions (an override-above is continued above, an override-below can no longer pull the seed down). The migration logs the seed result and warns when it exceeds ftReceiptNumerator, which is only possible via a historical caller-numbered submission. Test: UpgradedQueue_SeedsAtMaxOwnSeriesAa_NotAtNewest.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
StefanKert: Fixed in c32b3d1b: the SCU handwritten validation now rejects AA <= 0 (aligned with the queue’s inbound gate), with a comment on each side pointing at the other so they stay in sync. Tests: MapToInvoicesDoc_Handwritten_WithNonPositiveAA_ShouldReturnError for -5 and 0.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
StefanKert: Fixed in c32b3d1b: handwritten payloads whose Series ordinal-equals the queue’s InvoiceSeries are rejected before anything is submitted — numbering in that series is assigned exclusively by the middleware. Test: Handwritten_WithQueueOwnSeries_IsRejectedBeforeTheScu.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
StefanKert: Fixed in c32b3d1b: NumberStyles.None + CultureInfo.InvariantCulture, same as CountrySegment.TryParse.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
StefanKert: Acknowledged as the accepted-risk window (structure shared with the ES processors) — c32b3d1b documents it at the commit site, and CommitFailureAfterFiledInvoice_RetryReusesSameAa_Phase2Target pins the behavior as the red/green target for the 233 self-heal, which must treat this direction as “number consumed, advance”. Logging was skipped for now (the reservation has no logger); happy to plumb one through together with the Phase-2 work.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Models/Cases/SignatureTypeGR.cs
StefanKert: Pinned in c32b3d1b: MarkSignatureContract_IsPinnedForTheQueueCommitGate (scu-gr: enum value + ItemsChoiceType.invoiceMark caption) and MarkSignatureContract_IsPinnedAgainstScuGr (queue: enum value) — a drift now fails the build on whichever side moved. The warn-on-markless-success suggestion needs logger plumbing in the reservation and is deferred to the Phase-2 work.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEFactory.cs
StefanKert: Fixed in c32b3d1b — the ResolveSeriesAndAa comment and the AADEFactoryTests handwritten-test comment now describe the actual contract: the queue is the single writer of the segment (handwritten inbound, overrides rejected), nothing rewrites or compares afterwards, and filed/not-filed is decided by the invoiceMark gate.

queue/src/fiskaltrust.Middleware.Storage.AzureTableStorage/Repositories/Configuration/AzureTableStorageQueueGRRepository.cs
StefanKert: Fixed in c32b3d1b: LastInvoiceMoment now takes the queue-side ftReceiptMoment (always UTC) instead of the caller-supplied cbReceiptMoment.

queue/test/fiskaltrust.Middleware.Localization.QueueGR.UnitTest/Processors/InvoiceCounterReservationTests.cs
StefanKert: Added in c32b3d1b: SuccessfulSubmission_AdvancesInvoiceNumerator now asserts the persisted identification is exactly “ft1#CB-A-42”.

queue/test/fiskaltrust.Middleware.Localization.QueueGR.UnitTest/Processors/InvoiceCounterReservationTests.cs
StefanKert: Added in c32b3d1b: CommitFailureAfterFiledInvoice_RetryReusesSameAa_Phase2Target pins the stuck-retry behavior (the same aa is re-reserved on every retry after a filed-but-uncommitted invoice) as the Phase-2 target.

scu-gr/test/fiskaltrust.Middleware.SCU.GR.MyData.UnitTest/AADEFactoryTests.cs
StefanKert: Added in c32b3d1b: MyDataSCUProcessReceiptTests.ProcessReceiptAsync_OnSuccess_LeavesReceiptIdentificationUntouched drives the full SendInvoices success path against a stubbed HttpClient (new internal ctor seam + InternalsVisibleTo) and asserts the identification stays byte-identical and the mark signature is present.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEFactory.cs
StefanKert: Agreed — this goes into the PR description rewrite (pending, the description is stale on several counts by now) plus a release-note line along the lines of: “mydataoverride requests carrying invoiceHeader.series/aa are now rejected — override-based numbering no longer exists; caller-numbered paper documents use the HandWritten flag.” Leaving this thread open until the description is updated.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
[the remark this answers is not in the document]
StefanKert: Stale review (commit 219d6a547d) — this seeding branch no longer exists on the current head. The migration seeds from the queue-item submission history (maximum own-series aa found in successful responses), never from ftReceiptNumerator, and runs once at queue start. Fresh queues get InvoiceSeries seeded at activation and have no submission history, so they start at aa = 1 no matter how many NoOp receipts (closings, zero receipts) preceded the first myDATA submission — pinned by NewQueueWithNoOpsBeforeFirstSubmission_StartsAtAa1 and UpgradedQueue_NeverSubmitted_SeedsZero.

queue/src/fiskaltrust.Middleware.Storage.AzureTableStorage/Repositories/AzureTableStorageQueueItemRepository.cs
```diff
@@ -202,7 +202,11 @@ orderby queueItem.TimeStamp descending
 
         public async Task<ftQueueItem> GetByQueueRowAsync(long queueRow)
         {
-            var result = _tableClient.QueryAsync<TableEntity>(filter: TableClient.CreateQueryFilter<ftQueueItem>(x => x.ftQueueRow == queueRow));
+            // The partition key is derived from ftQueueRow (see MapToAzureEntity), so a
```
StefanKert: This looks kind of dangerous are we suer we want to do tht?

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/MyDataSCU.cs
StefanKert: Let's still write this back

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        if (!response.ReceiptResponse.ftState.IsState(State.Success) || TryExtractMark(response.ReceiptResponse) == null)
+        {
+            // Keep the segment only when AADE actually filed an invoice (Success plus
+            // invoiceMark). Failed submissions and SCU flows that never file one
+            // (delivery-note cancellation, payment methods) must persist exactly the
+            // identification they had before: the segment is the durable marker that
+            // this (series, aa) was consumed at AADE, and the queue-start migration
+            // seeds from it.
+            response.ReceiptResponse.ftReceiptIdentification = originalReceiptIdentification;
```
StefanKert, line 142: Is this behavior the same? So in case of not sending to mydata we use only the # before

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/MyDataSCU.cs
StefanKert: Restored in aa173d69: the SendInvoices success path writes the filed (series, aa) back into the segment again — an identity operation on queue-driven paths since the doc numbering derives from the very segment the queue stamped (pinned by ProcessReceiptAsync_OnSuccess_WritesBackAnIdenticalSegment), and direct SCU callers without a pre-appended segment get the historical append back (ProcessReceiptAsync_WithoutQueueSegment_AppendsFiledSeriesAndAa). The counter commit stays gated on the invoiceMark and does not depend on the write-back.

queue/src/fiskaltrust.Middleware.Storage.AzureTableStorage/Repositories/AzureTableStorageQueueItemRepository.cs
StefanKert: Good catch — it was genuinely dangerous as written: rows persisted before 49ce8fd5 (“fix queuerow hasing”, June 2024) were keyed by the plain row number, so a hashed-only partition query would have missed exactly the historical rows the counter migration walks. Fixed in aa173d69: the filter queries both key schemes (hashed + legacy plain row number), which stays index-backed either way and returns the same result set as the old property filter.

queue/src/fiskaltrust.Middleware.Localization.QueueGR/Processors/InvoiceCounterReservation.cs
StefanKert: Yes — identical to before. When nothing was filed at myDATA (error response, SCU exception, or an SCU flow that produces no invoice such as the payment-methods transmission), the queue restores the identification to exactly the pre-reservation value, i.e. bare “ft{N:X}#” — matching the old behaviour where the segment was only written on SendInvoices success. Pinned by ErrorResponse_DoesNotAdvanceCounter, SuccessWithoutMark_IsAnScuNoOp_DoesNotAdvanceCounter and SscdException_RestoresIdentification_AndDoesNotAdvanceCounter.

queue/src/fiskaltrust.Middleware.Storage.AzureTableStorage/Repositories/AzureTableStorageQueueItemRepository.cs
StefanKert: Update: reverted to the original property-filter implementation in fc0ae3e8 per author preference — the shared storage method stays untouched. The property filter finds rows regardless of their partition-key era; the migration walk’s per-row cost is a one-time trade-off at queue start.

PR #687, opened 2026-06-11
Missing transmission failure 2 in case of 5xx

AVPavel:
## Problem
The `Transmission_Failure2` signature was not being returned when a `5xx` error was received from the myDATA API.

## Description
* **New Method:** Created `AddTransmissionFailure2Signature` in `SignatureItemFactoryGR` to handle appending the Transmission Failure 2 signature.
* **New Enum:** Added a new member to `SignatureTypeGR`: `TransmissionFailure_2 = 0x4752_2000_0000_001C`.
* **Testing:** Added tests to replicate and verify behavior when receiving the following HTTP error codes: `500`, `502`, `503`, and `504`.

Fixes [#81](https://github.com/fiskaltrust/market-gr/issues/81)


StefanKert:
Closing, this is a fix that is needed in the GOV API

PR #705, opened 2026-07-08
docs: add RFC for single-currency enforcement in the Queue

StefanKert:
## Summary

Adds an RFC (`queue_single_currency`) specifying that a Queue processes and totalizes amounts in **exactly one currency**, determined by its market, and hard-validating the v2 `Currency` field against it so a queue can never accumulate more than one currency.

This is a **design document only** — no production code changes. It goes out for comments per the [RFC process](https://github.com/fiskaltrust/middleware/blob/main/rfcs/README.md).

📄 **Rendered:** [rfcs/0705-queue-single-currency.md](https://github.com/fiskaltrust/middleware/blob/claude/queue-multi-currency-rfc-026c0f/rfcs/0705-queue-single-currency.md)

## What & why

Currency handling in the queue is today ad-hoc and unspecified: most markets ignore `Currency` (effectively assuming EUR), while ES/PT hardcode an EUR-only rule inside their own localizations. The pattern that *should* hold everywhere — **a queue accepts exactly one, market-correct currency** — exists only as bespoke code in two markets.

This RFC **specifies** that single-currency rule and **lifts its enforcement into the shared `localization.v2` layer**, so every market (including non-EUR markets such as PL) enforces its correct currency the same way. It does **not** change the EUR-default behaviour, and the sum counters stay deliberately naïve — the rule just guarantees that only same-currency receipts are ever summed.

Key points:

- **Queue currency = f(market)** — hardcoded per market module; no `ftQueue` column, no config. "Version" enters only for a scheduled currency changeover, and means the middleware release version (`ProcessingVersion`), not the v1/v2 architecture.
- **Hard validation** — generalize the existing ES/PT `CurrencyMustBeEur` atom into a per-market `CurrencyMustMatchMarket`, wired through `MarketValidator`; plus charge-item / pay-item currency consistency.
- **Totalizers unchanged** — single-currency by construction, guaranteed by the validation.
- **No data-format change** — `Currency` already exists on `ReceiptRequest`/`ChargeItem`/`PayItem`.
- **Breaking** only for non-EUR markets, plus the narrow case of an EUR-market request that *explicitly* sends a non-EUR currency (ignored today, rejected after).
- **Changeover is optional** — the recommended path is simply to start a new queue.

## For reviewers

Open questions the RFC leaves for discussion: item-level currency strictness; whether to support the in-place (same-queue) changeover path at all vs. always mandating a new queue; and error-code stability for EUR markets.


rfcs/queue-multi-currency.md
```diff
@@ -0,0 +1,196 @@
...
+- AT: `ftQueueAT.ftCashTotalizer`
+- FR: the ~150 `*Totalizer` / `*CITotal*` / `*PITotal*` columns via [`QueueFR/Extensions/QueueFRExtensions.cs`](../queue/src/fiskaltrust.Middleware.Localization.QueueFR/Extensions/QueueFRExtensions.cs)
+
+They keep summing `ChargeItem.Amount` / `PayItem.Amount` as plain decimals. **Correctness now follows from validation**: because every accepted receipt is in the queue currency, every decimal added is in the same currency. The totalizer's currency is implicit (the queue currency), so we do **not** add a currency column. This should be documented at the accumulation sites: *"amounts are in the queue currency; single-currency is guaranteed by `CurrencyMustMatchMarket`."*
+
+## 4. Changeover mechanics
+
+`GetQueueCurrency(version)` is the single switch point. A changeover ships as **two transition versions** of the localization — current-currency and new-currency — with the switch at an effective version/date X; the recommended operational path is a fresh queue. No existing totalizer is ever retro-reinterpreted.
```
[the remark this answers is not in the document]
StefanKert: Good catch — fixed in 6a89bcf2. The changeover now requires a **fresh counter boundary at X**: a new queue is the recommended default, and the same-queue two-version path is explicitly invalid unless it resets/segments the totalizers at X (reusing FR's existing shift/day/month/year rollover), because otherwise the first post-X receipt is summed onto the old-currency total exactly as you describe. Reworked in both the "Currency changeover" section and Reference §4.

rfcs/queue-multi-currency.md
```diff
@@ -0,0 +1,196 @@
...
+# Guide-level explanation
+
+## The new concept: queue currency
+
+Every queue operates in exactly **one currency at a time**, fixed by *which market and localization version the queue is* — not something the POS picks per receipt, and not a setting an operator configures. That currency stays stable while the queue runs; the one exception is a deliberate, multi-step **currency changeover** (see [below](#currency-changeover)).
+
+- **EUR markets** (AT, DE, FR, IT, ES, PT, GR, …): the queue currency is EUR. Because `Currency` defaults to EUR and is *omitted from the JSON when it equals EUR*, existing POS integrations keep working with **no change at all**.
+- **Non-EUR markets** (PL → PLN, DK → DKK): the queue currency is the local currency. The POS **must** set `Currency` accordingly. If it omits the field, the value defaults to EUR, which does not match the queue currency, and the request is **rejected**.
+
+This is a breaking change **only for non-EUR markets** — which is defensible, because EUR markets are the default and the majority, and non-EUR markets are onboarded deliberately.
```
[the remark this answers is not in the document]
StefanKert: Agreed — scoped it in 6a89bcf2. The RFC now states that AT/DE/FR/IT/GR ignore `Currency` today, so a live request that explicitly sends a non-EUR `Currency` is accepted now but will be **rejected** once `CurrencyMustMatchMarket(EUR)` is wired in. Called out as the one EUR-market breaking case in the breaking-change note, the corner-cases list, and Drawbacks, with a sharpened rollout-audit item to find that traffic before the rule goes live. EUR/omitted-currency requests remain unaffected.

rfcs/0705-queue-multi-currency.md
```diff
@@ -0,0 +1,204 @@
...
+- **Rollout audit**: do any live EUR-market integrations (AT/DE/FR/IT/GR, which ignore `Currency` today) currently send an explicit non-EUR `Currency` that `CurrencyMustMatchMarket(EUR)` would suddenly reject? Audit and coordinate before wiring the rule in.
+
+# Future possibilities
+
+- **Foreign-currency payments** — paying a local-currency receipt with a foreign-currency tender. This would finish the DE `GetCurrency` / foreign-cash pay-item handling and the ME exchange-rate model, and needs FX rates and rounding rules. Distinct from queue currency.
+- **Multi-currency reporting / telemetry** — convert the queue currency to a reporting/base currency (`PosSystemMasterData.BaseCurrency`) for cross-market dashboards. A read-side concern, not queue semantics.
+- **Per-workstation queues** — raised in the same discussion in the context of per-register day-end (Kik / DeutscheFiskal / Swissbit). Orthogonal to currency, but it shares the "one queue = one well-defined scope" principle and warrants its own RFC.
```
steininger: remove customer and supplier names.

rfcs/0705-queue-multi-currency.md
```diff
@@ -0,0 +1,204 @@
...
+- **Canonical changeover path**: this RFC recommends a **new queue** as the default, because it is the only path that is totalizer-safe without extra reset logic. The same-queue two-version path stays available but must reset the totalizers at X. Confirm whether the same-queue path is worth supporting at all, or whether changeovers should simply mandate a new queue.
+- **Error-code stability**: keep `OnlyEuroCurrencySupported` verbatim for EUR markets (message / certification stability) or replace it everywhere with a generic `CurrencyMustMatchMarket`? Certification impact to confirm.
+- **Rollout audit**: do any live EUR-market integrations (AT/DE/FR/IT/GR, which ignore `Currency` today) currently send an explicit non-EUR `Currency` that `CurrencyMustMatchMarket(EUR)` would suddenly reject? Audit and coordinate before wiring the rule in.
+
+# Future possibilities
+
+- **Foreign-currency payments** — paying a local-currency receipt with a foreign-currency tender. This would finish the DE `GetCurrency` / foreign-cash pay-item handling and the ME exchange-rate model, and needs FX rates and rounding rules. Distinct from queue currency.
```
steininger: mention existing datastructure if beneficial.

rfcs/0705-queue-multi-currency.md
```diff
@@ -0,0 +1,204 @@
...
+- **Alternative B — a per-country default value for the `Currency` datatype.** Rejected. Changing the type's default per country breaks on a currency changeover and muddies serialization (omit-when-default is defined by the enum's zero being EUR). The type default stays EUR globally; correctness is enforced by validation instead.
+- **Alternative C — use `PosSystemMasterData.BaseCurrency`.** Rejected. It exists but is unused, configurable master data — same drift/misconfiguration risk as a queue field, and it is not version-aware for changeover.
+- **Alternative D — true multi-currency per item + FX conversion in totalizers.** Rejected / out of scope. The queue total is a technical receipt-chain sum, not accounting revenue. Multi-currency totals would require FX rates, a rounding policy, and per-currency columns (consider FR's already-150 totalizers × N currencies). It is not needed to serve a single-currency market.
```
steininger: maybe missleading.
also in accounting you define your book-currency. and this is equivalent like our market-lock on currency. 
any entries are in book currency, and if there is foreign currency accepted, than you need to proove the exchange-rate, by booking. 
as in point-of-sale world, you take in 100 usd and give the exchange in local currency, this exchange rate need to be recorded anyway on each single payitem.

steininger (changes requested):

rfcs/0705-queue-multi-currency.md
StefanKert: Done in 974bd407 — removed the customer and supplier names. The note now describes the per-register/per-workstation day-end scenario generically (a queue per workstation rather than one per outlet).

rfcs/0705-queue-multi-currency.md
StefanKert: Done in 974bd407 — the Foreign-currency payments item now points to the existing data-model hooks: `PayItem.Currency`, the `PayItemCaseFlags.ForeignCurrency` flag, DE's foreign-cash pay-item cases (`IsCashForeignCurrency` / the stubbed `GetCurrency`), ME's `CurrencyDetails { CurrencyCode, ExchangeRateToEuro }`, and BE's `ForeignCurrencyInput` — making the point that completing it is mostly wiring rather than new structures.

rfcs/0705-queue-multi-currency.md
StefanKert: Great framing — adopted it in 974bd407. Reworked Alternative D so it rejects only multi-currency *totalizers* (mixing within one total), **not** foreign-currency payments, and added your book-currency analogy: the market-locked queue currency is the accounting book currency. Foreign tender against a book-currency receipt is now explicitly legitimate and recorded as an exchange rate on each pay item (100 USD in, change in local currency), never mixed into the total. I also added the book-currency framing to the concept section and expanded the Foreign-currency payments item accordingly.

rfcs/0705-queue-multi-currency.md
```diff
@@ -0,0 +1,203 @@
+- Feature Name: `queue_multi_currency`
```
volllly: I think multi currency is a bad name for this concept 😅 It implies that multiple currencies can be used in parallel.

But it's actually about explicitly about making sure that only the one correct currency can be used in the queue. Also that only one currency per receipt can be used.
Also with the proposal for currency switchover this also makes sure that a queue only supports a _single_ currency at a time.

IMO this RFC is about supporting different/other/all/non-EUR currencies.
As RFCs are also meant to serve as documentation of the feature (kind of like ADRs) I'd get rid of the "multi" wording completely so we don't imply anything unintended. It's actually about the exact opposite of "multi" currency. It's about enforcing that it's _not_ multi currency.


```suggestion
- Feature Name: `queue_single_currency`
```
or
```suggestion
- Feature Name: `queue_local_currency
```

rfcs/0705-queue-multi-currency.md
```diff
@@ -0,0 +1,203 @@
...
+- Start Date: 2026-07-08
+- RFC PR: [fiskaltrust/middleware#705](https://github.com/fiskaltrust/middleware/pull/705)
+<!-- - Tracking Issue: [fiskaltrust/middleware#0000](https://github.com/fiskaltrust/middleware/issues/0000) -->
+- Markets: `PL`, `DK`, and every non-EUR market (EUR markets unaffected except requests sending an explicit non-EUR currency)
+
+# Summary
+
+A Queue processes and totalizes monetary amounts in a **single currency at a time** — the **queue currency** — fixed by the queue's **market and localization version**, not assumed to be EUR and not chosen per-receipt. It stays stable while the queue runs and changes only through a deliberate, multi-step currency changeover. The v2 `Currency` field (already present on `ReceiptRequest`, `ChargeItem` and `PayItem`) becomes authoritative and is **hard-validated**: a request whose currency does not match the queue currency is rejected, so mixed-currency receipts are impossible by construction. No change to the v2 data format is required; this is a middleware validation and totalizer-semantics change scoped to the v2 processing pipeline.
```
volllly: The "localization version" is mentioned multiple times in this RFC. But I'm not sure what is meant here by it.

When we talk about localization v1 or v2 we typically talk about the queues based on the different architectures `fiskaltrust.Middleware.Queue` (v1) or `fiskaltrust.Middleware.Localization.v2` (v2).

We also have the middleware version `1.3.xx` which is also stored s the `ProcessingVersion` in the queueitem.

volllly: > not assumed to be EUR and not chosen per-receipt.

From the rest of the RFC I gather that it _is_ infact assumed to be EUR because EUR is the default currency.

volllly: > changes only through a deliberate, multi-step currency changeover

It is not clear to me from the RFC if this is an optional or required feature.
Or if it is even a desired feature as below it says the recommended approach is to start a new queue.

rfcs/0705-queue-multi-currency.md
```diff
@@ -0,0 +1,203 @@
...
+
+# Motivation
+
+Non-EUR markets — Poland (PLN) is the immediate driver, Denmark (DKK) already has a queue table — need the Queue to operate in local currency. Today the middleware implicitly treats everything as EUR:
```
volllly: > Today the middleware implicitly treats everything as EUR:

As per this proposal it will continue to do so. It will just fail if EUR is the incorrect currency.

rfcs/0705-queue-multi-currency.md
```diff
@@ -0,0 +1,203 @@
...
+
+- Most markets (AT, DE, FR, IT, GR) **ignore** the `Currency` field entirely and silently assume EUR.
+- ES and PT **actively reject** anything that is not EUR (`CurrencyMustBeEur` → `OnlyEuroCurrencySupported`).
+- There is no country→currency configuration to build on.
```
volllly: There also will not be one when this RFC is implemented

rfcs/0705-queue-multi-currency.md
```diff
@@ -0,0 +1,203 @@
...
+
+A Polish POS therefore cannot be served correctly: either its amounts are silently mislabelled as EUR, or (in an ES/PT-style market) it is rejected outright.
```
volllly: IDK that's just a matter of convention. And the whole proposal builds of that fact that this is just convention.

> or (in an ES/PT-style market) it is rejected outright.

The ES/PT markets are already implementing the proposal by enforcing the correct currency. The PL queue would add a `CurrencyMustBePln` validation and be able to serve the markets.

---

This RFC is not about chanaging that behaviour it's about having a specification for it and about lifting the implementation of the spec from the market localizations to the general localization.v2. And an optional changeover mechanism.

rfcs/0705-queue-multi-currency.md
```diff
@@ -0,0 +1,203 @@
...
+
+The problem is sharpest at the **sum counters (Summenzähler)**. Every queue totalizer — the generic `ftQueue.ftReceiptTotalizer`, the AT `ftCashTotalizer`, and the ~150 per-rate/per-payment totalizers on `ftQueueFR` — is a plain single-currency `decimal` that blindly sums `ChargeItem.Amount`. These totals feed the receipt-chain hashes and country signatures. If two currencies ever entered one queue you would literally be adding PLN to EUR, and every signature built on that total would be meaningless. The counter must not be a naïve `10 € + 10 PLN + 10 USD = 30`.
```
volllly: > The counter must not be a naïve

The counter must and will strictly be and stay naïve because we don't want to have to handle multiple different currencies we don't want to have more complexity in the counters.

We want to to enforce in the localization.v2 that only receipts can be processed by the queue that we can sum up with the naïve counters.

rfcs/0705-queue-multi-currency.md
StefanKert: Agreed — the name implied the opposite of the intent. Renamed the feature to `queue_single_currency` (and the file to `0705-queue-single-currency.md`), and purged the word "multi" from the whole RFC. I went with *single* over *local* because the enforcement applies to every market, EUR included, not only the local-currency ones — happy to switch to `queue_local_currency` if you/the author prefer.

rfcs/0705-queue-multi-currency.md
StefanKert: Good point — clarified in §1: by *version* we mean the middleware release version (recorded as `ProcessingVersion` on the queue item), **not** the v1/v2 architecture. The queue currency is a function of the market alone; version only enters when a market schedules a currency changeover.

StefanKert: Correct — that phrasing was wrong. Removed "not assumed to be EUR"; the RFC now says the currency is determined by the market, and EUR remains the data-type default — which is exactly how EUR markets keep working unchanged.

StefanKert: Clarified — changeover is **optional** and the single-currency rule stands without it. The recommended path is simply to start a new queue (no new feature); whether we build any in-place same-queue mechanism at all is now flagged as an open question.

rfcs/0705-queue-multi-currency.md
StefanKert: Right — reframed. The middleware keeps EUR as the default; the RFC does not remove that, it adds enforcement so a wrong currency **fails** instead of being silently accepted. Motivation now says this explicitly.

rfcs/0705-queue-multi-currency.md
StefanKert: True — removed that bullet. The RFC deliberately does **not** add a country→currency config; the currency stays hardcoded in each market module. The motivation no longer implies a missing config to fill.

rfcs/0705-queue-multi-currency.md
StefanKert: This is the better framing and I adopted it throughout. Summary and Motivation now say the RFC **specifies** the single-currency rule and **lifts** enforcement out of the per-market localizations (ES/PT already do it) into shared `localization.v2`, plus an optional changeover mechanism — not a change to the behaviour itself. Dropped the "cannot be served / rejected outright" framing; PL just needs a `CurrencyMustBePln`, which is exactly the point being generalized.

rfcs/0705-queue-multi-currency.md
StefanKert: Agreed and reworded — the counters must and will stay naïve. The Motivation now frames the rule as *"only accept receipts we can sum naïvely"* (all one currency), so the naïveté of the counters is the thing we protect, not something we fix.

StefanKert:
All review feedback is addressed and the RFC is now a **single file** again: [`rfcs/0705-queue-single-currency.md`](https://github.com/fiskaltrust/middleware/blob/claude/queue-multi-currency-rfc-026c0f/rfcs/0705-queue-single-currency.md) (be3f311f). The `-restructured` and `-summary` companions are removed; their content is merged into the main document.

What the consolidated version carries from the review threads:

- **Naming** (@volllly): `queue_single_currency`, no "multi" wording anywhere.
- **"Version"** means the middleware release version (`ProcessingVersion`), not the v1/v2 architecture; the queue currency is a function of the market alone.
- **EUR default stays**: the RFC specifies the rule and lifts the ES/PT enforcement into shared `localization.v2`; it does not change today's behaviour, it makes a wrong currency fail instead of being silently accepted.
- **Counters stay naive** by design; the rule is "only accept receipts we can sum naively".
- **Changeover is optional** and in its own section: a new queue is the default path and needs no feature; an in-place switch is only valid with a totalizer reset at the switch point (@codex P1).
- **Breaking-change scope** (@codex P2): non-EUR markets, plus EUR-market requests that explicitly send a non-EUR currency, with a rollout-audit item.
- **Book-currency framing** and the FX-at-pay-item model (@steininger), with the existing data-structure hooks listed under Future possibilities; customer and supplier names removed.

Section layering follows the altitude feedback: Summary = what, Motivation = why, Guide = PosCreator examples with error responses, Reference = how, Rationale = why this design.

PR #706, opened 2026-07-08
Upgrade GR myDATA SCU schema from v1.0.12 to v2.0.1 (delivery-note types 9.1/9.2 + e-transport schemas)

### LINKED ISSUE #694: Upgrade GR myDATA SCU schema from v1.0.12 to v2.0.1 (adds delivery-note types 9.1/9.2 + e-transport API)
[v2.0.1 XSDs.zip](https://github.com/user-attachments/files/29300225/v2.0.1.XSDs.zip)

### Is your feature request related to a problem? Please describe.

The AADE myDATA schema bundled in the GR SCU (`scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/Scheme/`) is **v1.0.12**. It predates the digital delivery-note / e-transport ("Παραστατικά Διακίνησης") expansion, so several document types and the whole transport API surface are missing.

This was hit concretely in fiskaltrust/market-gr#243 — the **9.2 Collective Dispatch Note (Συγκεντρωτικό Δελτίο Αποστολής / ΣΔΑ)** cannot be produced, **not even via the invoice override**:

- `SimpleTypes-v1.0.12.xsd` enumerates `invoiceType` as … `8.6` → `9.3` → `11.1` — there is no `9.1` or `9.2`.
- The generated `InvoiceType` enum (`Scheme/response-v1_0_12.cs`) therefore has no `Item92`.
- `AADEFactory.ApplyInvoiceHeaderOverride` validates the override string against that enum and throws `ArgumentException: Invalid invoiceType override value '9.2'`.

So the schema version itself is the blocker.

### Describe the solution you'd like

Upgrade the bundled schema from **v1.0.12 → v2.0.1** (AADE preofficial XSD package — same family as the DeliveryNote v2.0.x PDF in market-gr#243; to be attached to this issue).

**1. Replace the XSDs** under `Scheme/` with the v2.0.1 set. New files beyond the current bundle (the e-transport surface):
`TransportTypes`, `RegisterTransfer`, `RejectDeliveryNote`, `ConfirmDeliveryOutcome`, `GetDeliveryStatusResponse`, `GenerateGroupQRCode` (+`Response`), `RequestGroupQRDetailsResponse`.

**2. Regenerate the C# model** (`Scheme/response-v1_0_12.cs` → `response-v2.0.1.cs`) via `xsd.exe` — the command is documented in `Scheme/README.md`. Note: `targetNamespace` is unchanged (`http://www.aade.gr/myDATA/invoice/v1.0`; only the schema `version` attribute bumps to `2.0.1`), so the generated C# namespace stays the same → low churn for existing serialization.

**3. Reconcile the enum/model consumers.** `invoiceType` gains **9.1, 9.2, 10.1, 10.2**; `InvoiceHeaderType` gains `toWeigh`. Touch points:
- `AADEMappings.RequiresCustomerInfo(InvoiceType)` — the `switch` **throws `NotSupportedException` for unmapped types**, so the new values must be handled.
- Review `AADEMappings.SupportsCounterpart` / `SupportsCorrelatedInvoices` / `SupportsMultipleConnectedMarks` / `SupportsIncomeClassification` for the new types.
- `AADEFactory.InvoiceTypeMap` is built via reflection over the enum, so the `invoiceType: "9.2"` override **starts working automatically** once `Item92` exists — this is what directly unblocks market-gr#243.
- Extend `MyDataOverrideTests` (`MapToInvoicesDoc_WithInvoiceTypeOverride_*` theory) with the new values.
- Optionally add a native `GetInvoiceType` mapping + override fields (e.g. `toWeigh`) if 9.2 should be selectable without an override.

### Additional context

- **Out of scope / follow-up:** actually *implementing* the new Delivery-Note REST endpoints introduced by v2.0.1 (`RegisterTransfer`→`Transport`, `RejectDeliveryNote`, `ConfirmDeliveryOutcome`, `GetDeliveryNoteStatus`, `GenerateGroupQRCode` + group-QR details). This issue is scoped to the **schema upgrade + regeneration + enum reconciliation** so the existing override path can emit 9.1/9.2. The new transport endpoints can be tracked separately.
- **`invoiceType` delta v1.0.12 → v2.0.1:** added `9.1`, `9.2`, `10.1`, `10.2` (51 → 55 values); none removed.
- Related: market-gr#243 (9.2 ΣΔΑ investigation), businesscase PR fiskaltrust/businesscase#388 (sample, currently `run: false` pending this upgrade).
- Code refs: `AADEMappings.GetInvoiceType` / `RequiresCustomerInfo`; `AADEFactory.InvoiceTypeMap` + `ApplyInvoiceHeaderOverride`; `Scheme/README.md` (regen command).

StefanKert:
## What & why

Upgrades the bundled AADE myDATA schema in the GR SCU (`scu-gr/…/Scheme/`) from **v1.0.12 → v2.0.1** and regenerates the C# model. The old schema predated the digital delivery-note / e-transport expansion, so `invoiceType` had no `9.1`/`9.2` — the `9.2` (Συγκεντρωτικό Δελτίο Αποστολής / ΣΔΑ) case from fiskaltrust/market-gr#243 could not be produced even via the override. After this PR the enum exists and the `invoiceType: "9.2"` override resolves automatically.

Closes #694.

## Changes

- **Schema swap:** replaced all `*-v1.0.12` XSDs with the 21 v2.0.1 XSDs (fixed a stray space in `GenerateGroupQRCodeResponse-v2.0.1.xsd`). The new e-transport endpoint schemas (`RegisterTransfer`, `RejectDeliveryNote`, `ConfirmDeliveryOutcome`, `GetDeliveryStatusResponse`, `GenerateGroupQRCode` + `Response`, `RequestGroupQRDetailsResponse`, `TransportTypes`) are bundled for reference; only `TransportTypes` is compiled (pulled in via `xs:include` from `InvoicesDoc`).
- **Model regen:** `response-v1_0_12.cs` → `response-v2.0.1.cs`, generated with **NETFX 4.8 `xsd.exe` (4.8.3928.0)**. I first regenerated the *v1.0.12* model with that exact tool and confirmed it reproduces the committed file byte-for-byte, so the model diff is only the real schema delta: `InvoiceType` +`9.1/9.2/10.1/10.2`, `InvoiceHeaderType` +`toWeigh`, `InvoiceRowType` +`movePurposeLine`/`otherMovePurposeLineTitle`, plus the additive transport types. `targetNamespace` is unchanged, so the C#/XML namespaces are stable.
- **Enum reconciliation** (`AADEMappings`): `RequiresCustomerInfo` no longer throws for unlisted invoice types — anything not explicitly listed (including the new `9.1/9.2/10.1/10.2`) falls back to `false`. The `invoiceType` override resolves `Item91/92/101/102` automatically via the reflection-built `InvoiceTypeOverrideMap`. Reviewed `SupportsCounterpart/CorrelatedInvoices/MultipleConnectedMarks/IncomeClassification` — their defaults already fit the new movement-doc types.
- **Override surface completeness:** audited every new v2.0.1 field against the passthrough override; added `toWeigh` to `InvoiceHeaderTypeOverride` and `movePurposeLine` / `otherMovePurposeLineTitle` to `InvoiceRowTypeOverride`, applied in `AADEFactory`. The remaining new fields are the out-of-scope e-transport surface or read-only myDATA values (see follow-ups).
- **Customer-info validation:** unchanged — still evaluated against the base receipt-case type (`GetInvoiceType`).
- **Tests:** extended the `invoiceType` override theory with `9.1/9.2/10.1/10.2`; added `toWeigh` and `movePurposeLine` override tests.
- Updated the Scheme regen `README.md` and two stale `"myDATA v1.0.12"` error strings.

## Two forced consequences of v2.0.1 (please note)

1. **`otherTransportDetails` removed.** v2.0.1 comments this element out of the invoice schema (multi-vehicle transport moved to the separate e-transport API). It was the *only* removed model member, and `AADEFactory` wrote to it. Removed the dead override path, `InvoiceOverride.OtherTransportDetails`, and `TransportDetailOverride.cs`. Callers still sending `otherTransportDetails` JSON are silently ignored (System.Text.Json skips unknown members); single-vehicle info remains available via `InvoiceHeaderTypeOverride.VehicleNumber`.
2. **`packingsDeclarations` `[XmlIgnore]`d.** `xsd.exe` flattens it into a jagged `PackagingDetailType[][]` that `XmlSerializer` cannot construct — it broke serialization of the *entire* `InvoicesDoc`. It is out-of-scope e-transport surface and never populated here, so it is excluded from (de)serialization (documented in the model + README as a required post-regen edit). Faithful support (a proper `PackingsDeclaration` wrapper) can come with the delivery-note API.

## Out of scope (follow-ups)

- Implementing the new delivery-note REST endpoints (RegisterTransfer/RejectDeliveryNote/ConfirmDeliveryOutcome/GetDeliveryNoteStatus/GenerateGroupQRCode + group-QR).
- A native `GetInvoiceType` mapping for 9.x/10.x (9.2 is override-selectable, which is what unblocks market-gr#243).
- Populating the new `ProviderSignatureType.EndToEndReferenceID` (IRIS payment reference) — tracked in fiskaltrust/market-gr#258.

## Verification

- Full `fiskaltrust.Middleware.SCU.GR.sln` builds with 0 errors.
- **823/823** MyData unit tests pass (including the new override cases and the XML-serialization round-trip test).


scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEMappings.cs
```diff
@@ -586,8 +586,15 @@ public static bool RequiresCustomerInfo(InvoiceType invoiceType)
...
+            // is issued before the recipient is known, so it does not require customer info.
+            // NOTE: today these are only reached defensively — GetInvoiceType never returns
+            // 9.1/9.2/10.1/10.2, and the invoiceType override is applied after validation —
+            // so the split has no runtime effect yet; it encodes the intended AADE semantics.
+            InvoiceType.Item11 or InvoiceType.Item12 or InvoiceType.Item13 or InvoiceType.Item14 or InvoiceType.Item15 or InvoiceType.Item16 or InvoiceType.Item21 or InvoiceType.Item22 or InvoiceType.Item23 or InvoiceType.Item24 or InvoiceType.Item51 or InvoiceType.Item52 or InvoiceType.Item31 or InvoiceType.Item32 or InvoiceType.Item61 or InvoiceType.Item62 or InvoiceType.Item71 or InvoiceType.Item81 or InvoiceType.Item91 or InvoiceType.Item93 or InvoiceType.Item101 or InvoiceType.Item102 => true,
+            InvoiceType.Item82 or InvoiceType.Item84 or InvoiceType.Item86 or InvoiceType.Item85 or InvoiceType.Item92 or InvoiceType.Item111 or InvoiceType.Item112 or InvoiceType.Item113 or InvoiceType.Item114 or InvoiceType.Item115 => false,
```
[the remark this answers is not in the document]
StefanKert: Fixed in 96f214e7. Validation now evaluates the customer-info requirement against the **effective** invoice type: `ValidationGR` requires customer info only when *both* the base receipt-case type and the overridden type require it. So a delivery-note+transport case (→ 9.3) overridden to `9.2` (ΣΔΑ) now validates without a customer, while an override to `9.1` (which still identifies a recipient) still fails `CustomerInfoRequired`. Added `AADEMappings.GetEffectiveInvoiceType(request)` and regression tests for both directions.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/Validation/ValidationGR.cs
```diff
@@ -28,7 +28,15 @@ public static (bool, MiddlewareValidationError? middlewareValidationError) Valid
...
         }
 
-        if (AADEMappings.RequiresCustomerInfo(AADEMappings.GetInvoiceType(receiptRequest)) && !receiptRequest.ContainsCustomerInfo())
+        // Customer-info requirement is driven by the receipt-case type, but an invoiceType
+        // override can RELAX it: e.g. overriding a delivery-note case (which maps to 9.3 and
+        // normally requires customer info) to 9.2 (ΣΔΑ / collective dispatch) — which has no
+        // determined recipient — must be allowed without a customer. Require customer info only
+        // when both the base type and the effective (possibly overridden) type require it, so an
+        // override can waive the requirement but never bypasses it for a type that still needs it.
+        if (AADEMappings.RequiresCustomerInfo(AADEMappings.GetInvoiceType(receiptRequest))
```
StefanKert: Let's bring the same  behavior as earlier back. For now we'll rather not break sth

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEMappings.cs
```diff
@@ -586,8 +621,15 @@ public static bool RequiresCustomerInfo(InvoiceType invoiceType)
...
             _ => throw new NotSupportedException($"The invoice type '{invoiceType.GetXmlEnumAttributeValueFromEnum()}' is not supported"),
```
StefanKert: If a invoice type is not covered above we simply fall back to false

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/Validation/ValidationGR.cs
StefanKert: Reverted in 32ef7ad2 — `ValidationGR` is back to the original check against the base receipt-case type (`RequiresCustomerInfo(GetInvoiceType(...))`); the override-aware/effective-type rule and the now-unused `GetEffectiveInvoiceType` helper are removed. Behavior matches earlier.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEMappings.cs
StefanKert: Done in 32ef7ad2 — the default arm now returns `false` instead of throwing, so any invoice type not explicitly listed (including the new v2.0.1 delivery-note types 9.1/9.2/10.1/10.2) falls back to `false`. Restored the original explicit true/false lists.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEMappings.cs
```diff
@@ -588,7 +603,9 @@ public static bool RequiresCustomerInfo(InvoiceType invoiceType)
         {
             InvoiceType.Item11 or InvoiceType.Item12 or InvoiceType.Item13 or InvoiceType.Item14 or InvoiceType.Item15 or InvoiceType.Item16 or InvoiceType.Item21 or InvoiceType.Item22 or InvoiceType.Item23 or InvoiceType.Item24 or InvoiceType.Item51 or InvoiceType.Item52 or InvoiceType.Item31 or InvoiceType.Item32 or InvoiceType.Item61 or InvoiceType.Item62 or InvoiceType.Item71 or InvoiceType.Item81 or InvoiceType.Item93 => true,
             InvoiceType.Item82 or InvoiceType.Item84 or InvoiceType.Item86 or InvoiceType.Item85 or InvoiceType.Item111 or InvoiceType.Item112 or InvoiceType.Item113 or InvoiceType.Item114 or InvoiceType.Item115 => false,
-            _ => throw new NotSupportedException($"The invoice type '{invoiceType.GetXmlEnumAttributeValueFromEnum()}' is not supported"),
+            // Any invoice type not listed above (incl. the v2.0.1 delivery-note types
+            // 9.1/9.2/10.1/10.2) falls back to false rather than throwing.
+            _ => false,
```
StefanKert, line 608: Two notes on this switch:

**1. Validation runs on the base type, before the override.** `RequiresCustomerInfo` is called only from `ValidationGR.cs:31` as `RequiresCustomerInfo(GetInvoiceType(request))` — the base type — and `ValidationGR` runs (`AADEFactory.cs:142`) *before* the `invoiceType` override is applied (`AADEFactory.cs:147`). So a base-9.3 delivery note overridden to `9.2` (ΣΔΑ, no customer) is still rejected with `CustomerInfoRequired` before the override runs. This is the main open question — see the review summary.

**2. `_ => false` drops the old `throw` guard.** Inert today (every type `GetInvoiceType` can return is explicitly listed above), but if `GetInvoiceType` is later extended to return a new type that legally requires customer info, this silently accepts it without customer info instead of failing loud. Fine as a deliberate choice — just flagging the trade-off; a one-line comment noting the intent would help the next reader.

scu-gr/test/fiskaltrust.Middleware.SCU.GR.MyData.UnitTest/MyDataOverrideTests.cs
```diff
@@ -1114,7 +1150,11 @@ public void MapToInvoicesDoc_WithInvalidCurrencyOverride_ShouldReturnError()
     [InlineData("8.4", InvoiceType.Item84)]
     [InlineData("8.5", InvoiceType.Item85)]
     [InlineData("8.6", InvoiceType.Item86)]
+    [InlineData("9.1", InvoiceType.Item91)]
+    [InlineData("9.2", InvoiceType.Item92)]
```
StefanKert, line 1154: These cover the factory's override **resolution** (string → `InvoiceType`), which is good — but nothing exercises the `ValidationGR` gate that decides whether a `9.2` receipt actually survives. A test that runs a **customer-less, base-9.3 delivery note with an `invoiceType: "9.2"` override through `MapToInvoicesDoc`** would pin the intended end-to-end behavior (pass or fail) and prevent silent regressions in either direction. See the review summary for context.

StefanKert (commented):
## Automated review — GR myDATA schema v1.0.12 → v2.0.1

Built the SCU (0 errors) and ran the MyData unit tests (**823/823 pass**). The schema swap, model regen, `toWeigh` / `movePurposeLine` overrides, `otherTransportDetails` removal, and the `packingsDeclarations` `[XmlIgnore]` are all clean and well-documented (model + README). One substantive issue that affects whether the PR meets its stated goal, plus two smaller notes inline.

### ⚠️ The `9.2` (ΣΔΑ) override is still blocked at validation for the no-customer case

`MapToInvoicesDoc` runs `ValidationGR.ValidateReceiptRequest` (`AADEFactory.cs:142`) **before** the `invoiceType` override is applied in `CreateInvoiceDocType` (`AADEFactory.cs:147`). `ValidationGR.cs:31` checks `RequiresCustomerInfo(GetInvoiceType(request))` against the **base** type. A delivery-note + transport receipt maps to `Item93` (9.3) → `RequiresCustomerInfo == true`. ΣΔΑ (9.2) is collective dispatch with recipient unknown → no customer info, so validation throws `CustomerInfoRequired` before the override ever runs.

Commit 96f214e7 identified and fixed exactly this; commit 32ef7ad2 reverted it, so `ValidationGR.cs` is unchanged vs `main` and the block is back. The PR description still states the `invoiceType: "9.2"` override "resolves automatically" / "unblocks market-gr#243" — that is true only for enum resolution inside the factory, not end-to-end.

**Please confirm the target market-gr#243 receipt:**
- if it carries customer info → it works today (but "ΣΔΑ with a known recipient" is a compliance question for the GR team);
- if it is customer-less → still blocked, and the description should be corrected to say 9.2 is now *override-selectable* while the no-customer ΣΔΑ path awaits a follow-up validation change (deferred per review).

Recommendation: don't re-litigate the validation revert here — correct the description and add the guard test noted inline. The schema + enum + factory resolution is real value on its own.

See the two inline comments for the smaller items.

PR #708, opened 2026-07-13
Fix 103 110 gr discounts

AVPavel:
Compliant GR discount mapping: fold discounts into `netValue`, distribute a whole-basket discount across lines, and emit `recType=6` for 100%-discount zero lines — instead of misusing `deductionsAmount`.

### Problem
The GR SCU mapped every `ExtraOrDiscount` charge-item modifier to the myDATA `deductionsAmount` field and left the line's `netValue` at its pre-discount value. `deductionsAmount` is a statutory **withholding/retention** (Κρατήσεις), not a commercial discount — causing two compliance defects, both reproduced on the myDATA dev sandbox:

- **Partial discounts (market-gr#103):** accepted by AADE but **not compliant** — reported income was overstated (the `incomeClassification` amount stayed at the pre-discount net) and the withholdings field was misused.
- **100% discounts (market-gr#110):** **rejected** by AADE rule 241 (`deductionsAmount` may not exceed the line `netValue`) — impossible to satisfy at 100% off.

### Change
- **Fold discounts into the line.** An `ExtraOrDiscount` modifier is folded into the parent line's `netValue`/`vatAmount` (VAT recomputed from the post-discount gross, half-away-from-zero), `discountOption=true` is set, the income/expenses classification amount tracks the post-discount net, and **no `deductionsAmount` is emitted**. Refund sign handling is preserved. (`AADEFactory`)
- **Zero-net lines.** When a line nets to zero on an invoice type that enforces myDATA rule 222 (1.1), it is emitted with `recType=6` and no classification; retail 11.x keeps its normal classification. New helper `AADEMappings.EnforcesZeroLineRecType`.
- **Whole-basket discount (market-gr#103, abstracted entry point).** A rate-agnostic discount (`ExtraOrDiscount` + `VATRate 0` + `Position 0`) is distributed across all base lines proportionally **by gross** (rounding remainder to the largest line), then folded per line at each line's own VAT rate — so a single discount line yields a compliant, distributed document across mixed VAT rates.
- **Guard.** A discount exceeding the line it applies to (or a basket discount exceeding the basket total) is rejected with a clear error instead of producing an AADE-invalid document.

### Tests
`GrDiscountBusinessCaseTests` mirrors the nine merged businesscase examples in `example/SignRequestReceipt_GR_Discounts` (per-line uniform / mixed-VAT / B2B, abstracted basket uniform + mixed, 100% B2B via modifier, direct zero line, retail-zero regression guard, over-the-line anti-pattern).

Fixes fiskaltrust/market-gr#103
Fixes fiskaltrust/market-gr#110

### DIFF
--- a/scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEFactory.cs
+++ b/scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEFactory.cs
@@ -180,8 +180,9 @@ public class AADEFactory
             return;
         }
 
-        var baseItems = items.Where(ci => !ci.IsDiscountOrExtra() && !ci.IsVoucherRedeem()).ToList();
-        var totalBaseGross = baseItems.Sum(ci => ci.Amount);
+        var grouped = receiptRequest.GetGroupedChargeItems();
+        var baseLines = grouped.Select(g => g.chargeItem).Where(c => !IsBasketDiscount(c)).ToList();
+        var totalBaseGross = baseLines.Sum(c => c.Amount);
         if (totalBaseGross <= 0m)
         {
             throw new Exception("A whole-basket discount requires base charge items with a positive total amount.");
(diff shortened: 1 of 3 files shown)


AVPavel:
added release notes label for the whole basket discount -> the way it should be sent needs to be available for every integrator.

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEFactory.cs
```diff
@@ -160,6 +163,76 @@ public static bool HasInvoiceTypeOverride(ReceiptRequest receiptRequest)
         }
     }
 
+    // A whole-basket (rate-agnostic) discount is signalled by the ExtraOrDiscount flag together
+    // with VATRate 0 and Position 0 (case 0x..._0004_0010). It belongs to no single line, so it is
+    // distributed across all base lines by gross rather than attached to one line.
+    private static bool IsBasketDiscount(ChargeItem ci)
```
StefanKert, line 169: We should have a method already for grouping items with discounts afaik. This gives us a collection of chargitems with their connected modifiers

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEFactory.cs
```diff
@@ -1337,8 +1400,66 @@ private static List<InvoiceRowType> GetInvoiceDetails(ReceiptRequest receiptRequ
             }
             if (grouped.modifiers.Count > 0)
             {
-                invoiceRow.deductionsAmount = grouped.modifiers.Sum(x => x.Amount) * -1;
-                invoiceRow.deductionsAmountSpecified = true;
+                // A discount/extra modifier is folded into the line's netValue/vatAmount — it is
+                // NOT reported as deductionsAmount (a statutory withholding; cf. AADE rule 241).
+                var modGrossSum = grouped.modifiers.Sum(m => m.Amount);
```
StefanKert, line 1405: Can we add some Unittests for thsi logic? Happy with the itnegrationtests, but adding them could make sense

scu-gr/src/fiskaltrust.Middleware.SCU.GR.MyData/AADEFactory.cs
```diff
@@ -1337,8 +1400,66 @@ private static List<InvoiceRowType> GetInvoiceDetails(ReceiptRequest receiptRequ
...
+                if (invoiceRow.expensesClassification != null)
+                {
+                    foreach (var ec in invoiceRow.expensesClassification)
+                    {
+                        ec.amount = invoiceRow.netValue;
+                    }
+                }
+            }
+
+            // A resulting zero-net line on an invoice type that enforces rule 222 (e.g. 1.1) must
```
StefanKert, line 1453: Do we have documentation on this, and a reference why this is fine? This seems like a odd case to me so let's make sure that this is documented properly

AVPavel: Sending a zero-net line on a 1.1 invoice without recType=6 is rejected by myDATA with this message: "NetValue per line, which have recType != 6, must have value greater than 0 for this invoice type." Check [this](https://developer.fiskaltrust.eu/#/pos-system/GR?endpoint=sign&businesscase=SignRequestReceipt_GR_Discounts_110_B2BInvoice_ZeroLine_Direct)

I can extend the documentation at the source code level if you wish. It is also documented at the businesscase repository level

PR #710, opened 2026-07-17
[RFC] Automatic SCU switch

volllly:
In this RFC we introduce the possibility to configure an SCU switch to be performed automatically by the queue during daily closing.

[Rendered](https://github.com/fiskaltrust/middleware/blob/rfc/automatic-scu-switch/rfcs/0710-automatic-scu-switch.md)

### DIFF
--- a/rfcs/0710-automatic-scu-switch.md
+++ b/rfcs/0710-automatic-scu-switch.md
@@ -39,18 +39,18 @@ Because of this when switching to a cloud TSE we can automatically perform the S
 When switching to a cloud TSE we can already check if the target SCU is reachable before starting the process with the init-SCU-switch.
 If the target SCU is not reachable we can abort the process before it starts.
 (With the manual SCU switch process we have to first perform the init-SCU-switch before we can check if the target SCU is reachable.
-If the switch is incorrectly configured or the target SCU broken this sometimes leads to broken states that _have_ to be reset with a void init-SCU-switch receipt.
-We can catch those cases before they occur with the automatic switch)
+If the switch is incorrectly configured or the target SCU is broken this sometimes leads to broken states that _have_ to be reset with a void init-SCU-switch receipt.
+We can catch those cases before they occur with the automatic switch.)
 
-The process this RFC describes, makes it possible to automatically switch to cloud SCUs during daily closing.
+The process this RFC describes makes it possible to automatically switch to cloud SCUs during daily closing.
 
 This automation enables us to:
-- Bundle the error prone manual SCU switch process into a single receipt that automatically handles all errors
+- Bundle the error-prone manual SCU switch process into a single receipt that automatically handles all errors
 - Perform SCU switches where the only PosDealer interaction happens in the portal. No input by the PosOperator or PosSystem is needed.
 - Allow our support to switch queues over to a new working SCU (e.g. if a cloud TSE outage leads to broken/deleted TSEs) without PosDealer interaction and without the need for the PosSystem to implement the SCU-switch process.
 
 The automatic SCU switch reuses all of the logic and processes from the init-SCU-switch and finish-SCU-switch receipts.  
-Because of this we don't need to think about legal implications or downstream processes.
+Because of this we don't need to re-evaluate the legal implications of the individual steps or downstream processes.
 We know that the init-SCU-switch works and we know that the finish-SCU-switch works. That's why we can execute them automatically and bundle them into the daily closing receipt.
 (In the manual process the daily-closing receipt also needs to happen immediately before the init-SCU-switch receipt is sent).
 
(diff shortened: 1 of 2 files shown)


rfcs/0710-automatic-scu-switch.md
```diff
@@ -0,0 +1,210 @@
...
+The SCU switch currently is a five-step process:
+1. Configure the SCU switch with source SCU and target SCU
+2. Perform a daily-closing receipt
+3. Perform the init-SCU-switch receipt
+4. For hardware TSEs: Switch out the old TSE device for the new TSE device
+5. Perform the finish-SCU-switch receipt
+
+</details>
+
+The current SCU switch process is designed that way because for hardware TSEs you need to physically switch the TSE devices out in the middle of the process.
```
forsthug: be aware that there could be multiple switches/migrations and all must be visible in dsfinvk, possible queue of scu with start end - source and target are no longer enough.

rfcs/0710-automatic-scu-switch.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Implement an all-in-one-SCU-switch receipt
+
+Instead of configuring it to happen automatically and doing it automatically during daily closing, we could also create a new receipt case that does both init- and finish-SCU-switch in one go.
+
+This has the advantage of being explicit but the disadvantage of requiring a specific receipt to be sent.
+
+# Unresolved questions
+
+- Do we need another parameter in the `ModeConfigurationJson` to specify that the automatic init-SCU-switch should be done with the force flag?
```
StefanKert: One of the options that we have is covering this as some sort of "error" case similar to when the Operator just plugs in another TSE. Specifically for hardware TSEs it is today possible that people just switch it. This is something that we'll see often in the upcoming months, specifically if TSEs are retiring and the replacement process is not sound.

As an Unresolved question I would add, if it makes sense to connect these processes into a single one considering the "Error Switch" case, where the TSE does not really match with what we are expecting to see and handle it in a similar ways as other Failure Moes with a zeroreceipt

rfcs/0710-automatic-scu-switch.md
```diff
@@ -0,0 +1,210 @@
...
+
+When opening the Queue -> SCU connection dialog and configuring the SCU switch, the portal detects if the SCU switch could be performed automatically and gives the user the option to select the automatic SCU switch.
+
+If the user opts for the automatic SCU switch, it will be performed automatically on the next daily closing after the cashbox has been rebuilt and the queue has received the new configuration.
+
+If the user does not opt for the automatic SCU switch, the current behaviour (manual SCU switch has to be performed) is kept.
+
+### TSE exchange workflow
+
+The TSE exchange workflow in the portal also detects if the SCU switch can be performed automatically and, if so, configures the cashbox for the automatic SCU switch and informs the user that no further action is needed.
```
StefanKert, line 152: @Tjark-fiskaltrust  @forsthug I think that we need t add a dedicated section here that covers the legal point of view.

Specifically in relation to the following questions:

1. How does the Market handle multiple TSEs in a single day, will they be exported in a single DSFinVK?
2. Is there a mandate that prevents us from having multiple TSEs durign a day?
3. How does this affect the Notifications that need to be done. If we "remove" the old TSE, do we need to deregister it, or is this more ont he level of a Queue / AUfzeichnungssystem

rfcs/0710-automatic-scu-switch.md
volllly: The automatic switch will do all of the same steps including actionjournals etc. as the manual switch so everything that works for the manual switch should also work for the automatic switch and if the manual switch is compliant the automatic switch should be aswell.

The only difference is that this is bundled into one receipt. Do you think we'll get unsolvable problems in the DSFinV-K when when both init and finish are done in the same receipt?

rfcs/0710-automatic-scu-switch.md
volllly: Just for clarification, with the automatic process there will not be more TSEs used per day than before.

Manual Process: 
1. daily-closing with source TSE
2. init with source TSE
3. finish with target TSE
4. some normal receipts during the day
5. daily closing with target TSE

So during the day that the daily closing with the target TSE closes we have signatures with the target TSE but also have the one init signature with the source TSE.

Automatic Process:
1. daily-closing with source TSE
1.1. Automatic init with source TSE
1.2. Automatic finish with target TSE
2. some normal receipts during the day
3. daily closing with target TSE

So in this case in the day of the first daily closing we have signatures with the source TSE and the one finish signature of the target TSE.

---

So either the signature of the init with the source TSE or the signature of the finish with the target TSE are in the "wrong" day.

rfcs/0710-automatic-scu-switch.md
```diff
@@ -0,0 +1,210 @@
...
+    }
+  ]
+}
+```
+
+</details>
+
+## Trigger of the automatic SCU switch
+
+After (successful or unsuccessful) processing of the daily-closing receipt and before returning the ReceiptResponse, the queue checks if the `ModeConfigurationJson` of the current SCU has `PerformSwitchOnDailyClosing` set to `true`,
```
StefanKert: What happens to the TAR file on the old TSE?

volllly: If the daily closing is unsuccessful we implicitly use the force flag on the init-SCU-switch receipt.

We frontload the decision if the switch should be done or not into the configuration step.
There we can warn users of the implications of doing this if the SCU is broken/out of order.

So like with the manual flow with the force flag the TAR-File stays on the SCU/gets lost.

---

If we want to have more control over if the automatic switch should be forced we can include something like a `ForceSwitchOnDailyClosing` parameter to the `ModeConfigurationJson` (See unresolved questions).
You then get more control of the behaviour during configuration of the switch. If you know your source SCU is working you can configure the automatic switch to not force the init. If you know your source SCU is not working you can configure the automatic switch to force the init.

PR #711, opened 2026-07-17
fix(epsonrtserver): make SCU cloud-safe on the stateless CloudCashbox host

accodev:
## Problem

The `EpsonRTServer` SCU crashes at init on the hosted **CloudCashbox** platform:

```
System.ArgumentException: The value cannot be an empty string. (Parameter 'path')
  at System.IO.Directory.CreateDirectory(String path)
  at EpsonRTServerSCU.PersistState()
```

`_scuCacheFolder` falls back to `Environment.GetFolderPath(SpecialFolder.Personal)`, which returns `""` in the Linux container (no `$HOME`), and `PersistState()` called `Directory.CreateDirectory("")`.

## Why "best-effort", not a state store

Validated against the *RT Server Security Communication Protocol Rev27* + *Fiscal ePOS Metadata Guide Rev17*:

- The **device is authoritative** — it stores the whole per-till chain (last `CCDC/Hash/Token`, next receipt number, Z, daily amount). `createToken` re-anchors all of it and is the sanctioned recovery. `CCDC` is plain SHA-256 — no client-only secret. So the local state file is a **pure cache**; losing it loses nothing.
- **Single-writer per till is already guaranteed** by CloudCashbox's per-cashbox `Medallion.Threading` Redis sign-lock (one cashbox → one queue → one tillId) + header-based routing.

So the fix makes local persistence best-effort rather than adding any state-store abstraction, flag, or shared store.

## What changed

- **`EpsonRTServerSCU`** — `PersistState`/`LoadStateFromDisk` guard on an empty/unwritable folder (best-effort, latched); on the stateless host the till state is in-memory and re-seeded from the device via `createToken`. On-prem behavior is unchanged.
- **`EpsonRTServerCommunicationQueue`** — when no writable cache folder exists: skip directory creation, do not start (or restart, on daily closing) the background drain, and **force synchronous send** even if `SendReceiptsSync == false`. This prevents silent loss of fiscal documents to ephemeral disk — the one genuinely dangerous behavior.
- A tiny test-only seam (`Func<string>? personalFolderProvider = null`) makes the no-folder path deterministically testable; production callers pass nothing.

## Testing

`dotnet test scu-it/test/fiskaltrust.Middleware.SCU.IT.EpsonRTServer.UnitTest` → **63/63 passing**. New tests: no-crash + re-seed-once with no folder, force-sync with no folder, and the daily-closing drain guard (RED→GREEN).

## Cross-repo / deployment

Host registration (add `EpsonRTServer` to `ScuPackages` + the factory switch) is a **separate PR in `service-cloudcashbox`** (branch `epsonrtserver-scu-registration`). It only takes effect once that repo's `middleware` submodule pointer is bumped to a commit containing this PR.

## Follow-up (separate ticket)

`CustomRTServer` has the same bug class **live today** — its `SendReceiptsSync` defaults to `false` and buffers fiscal documents to an ephemeral relative path in cloud. Recommend prioritizing an audit / the same force-sync guard.

### DIFF
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerCommunicationQueue.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerCommunicationQueue.cs
@@ -22,6 +22,7 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTServer
         private readonly ILogger<EpsonRTServerCommunicationQueue> _logger;
         private readonly EpsonRTServerConfiguration _configuration;
         private readonly string _documentsPath;
+        private readonly bool _diskCacheAvailable;
 
         private bool _requestCancellation;
         private bool _processingReceipts;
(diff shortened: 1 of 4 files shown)


accodev:
/deploy scu-it EpsonRTServer
[](https://github.com/fiskaltrust/middleware/actions/runs/29594365124)

StefanKert (approved):

PR #714, opened 2026-07-21
fix(epsonrtserver): force synchronous signing on stateless hosts (LauncherEnvironment)

accodev:
## Problem

The "async only when a writable folder exists" guard (from #711) is an unreliable foundation for a fiscal invariant. `Environment.GetFolderPath(SpecialFolder.Personal)` is **environment-dependent** on Linux:
- On the CloudCashbox image (`aspnet:8.0-bookworm-slim`, root, no `$HOME`) it resolves to **`/root`** (non-empty) → `_diskCacheAvailable = true` → async is **not** forced. With `SendReceiptsSync=false`, fiscal documents buffer to the ephemeral `/root` and are **lost on pod restart** (no PVC, 2 replicas).
- In the original incident container it resolved to `""` (hence the crash we fixed) — proving the value varies by environment.

So async signing must not depend on inferring "cloud" from the disk.

## Change

- Add `EpsonRTServerConfiguration.LauncherEnvironment` (host-injected).
- `EpsonRTServerCommunicationQueue` treats `LauncherEnvironment == "cloud"` (case-insensitive, trimmed) as **no durable cache**, AND-ing it into `_diskCacheAvailable`. This single change flows through every existing guard → forced synchronous send, no disk buffering, no background drain.
- Warns (never hard-fails) when `SendReceiptsSync=false` is overridden.

## The rule

Asynchronous (offline-buffered) signing is enabled **iff** `SendReceiptsSync == false` **AND** `LauncherEnvironment != "cloud"` **AND** a cache folder resolves — otherwise **synchronous**. Async is for **on-prem/local (persistent disk) only**.

## Requires companion change

This is **inert until the host injects `LauncherEnvironment=cloud`** into the SCU's config — see the companion `service-cloudcashbox` PR. Until then it safely defaults to today's behavior. (The queue already receives `LauncherEnvironment=cloud`; the SCU config did not — the companion PR fixes that.)

## Tests

New test proves the invariant: a real `ServiceFolder` is present (so the folder heuristic alone *would* allow async) + `SendReceiptsSync=false` + `LauncherEnvironment=cloud` → sync path taken (`CreateReceiptAsync` called, non-null response) and **zero** files written under the folder. Full suite: **64/64**.

### DIFF
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerCommunicationQueue.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerCommunicationQueue.cs
@@ -47,12 +47,25 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTServer
                 _documentsPath = configuration.CacheDirectory!;
             }
 
-            _diskCacheAvailable = !string.IsNullOrEmpty(configuration.CacheDirectory) || !string.IsNullOrEmpty(cacheFolder);
+            // Host-injected signal (set by the launcher on CloudCashbox): a folder resolving on a stateless host
+            // (e.g. Environment.SpecialFolder.Personal -> "/root" on the cloud image, root, no $HOME) still looks
+            // writable, so the folder heuristic alone cannot be trusted there. LauncherEnvironment="cloud"
+            // overrides it explicitly.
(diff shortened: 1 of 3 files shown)


accodev:
Superseded by #716 — a leaner design: instead of a new LauncherEnvironment config property + cross-repo injection + submodule-bump lockstep, the SCU simply requires an explicit durable folder for async and drops the fragile SpecialFolder.Personal fallback. Closing in favor of #716.

PR #720, opened 2026-07-22
fix(epsonrtserver): record local emission time from the UTC receipt moment

accodev:
## Problem
`cbReceiptMoment` is UTC, but the RT Server records the **local** emission time. The SCU sent `cbReceiptMoment` verbatim as the device `dateTime` and the `<rt-doc-moment>` signature while separately declaring `srtUtcOffset` — so the recorded/printed fiscal time was UTC (e.g. 12:00) instead of local (14:00 in summer), and near midnight the **date** could be off by a day.

## Spec (verified in the Metadata Guide Rev17, §3.8)
- `dateTime` = "Fiscal receipt emission date and time (YYYYMMDDThhmmss format)" — local wall time.
- `srtUtcOffset` = "The time offset from UTC. Assumes values 1 (winter) or 2 (summer)".

## Fix
Convert to the device's local time at the single point the moment enters the fiscal document (`EpsonRTServerMapping.BuildFiscalDocument`), using the **device-reported `srtUtcOffset`** — correct on **any** host (the cloud host runs in UTC, so `ToLocalTime()` would be wrong; the device offset is host-independent). EpsonRTServer-only; the shared `SignatureFactory` format is unchanged.

`cbReceiptMoment` is UTC **by contract** ([fiskaltrust interface-doc](https://github.com/fiskaltrust/interface-doc/blob/master/doc/general/data-structures/data-structures.md): "Must be provided in UTC") but the `DateTimeKind` is **not** guaranteed at runtime — QueuePT explicitly rejects `Kind != Utc` (`ReceiptRequestValidations.cs:160`), and IT has no such guard. `ToRtServerLocalTime` therefore normalizes by kind — `Local` → `ToUniversalTime`, zoneless (`Unspecified`) → honored as UTC per the contract — before applying the offset.

## Refunds
- **Referenced** refunds/voids source the moment from the response signatures (`RTReferenceDocumentMoment`, injected by the queue from the original receipt) — unchanged.
- **Unreferenced** refunds/voids no longer fabricate a date from `cbReceiptMoment` (~today): `referenceDocMoment` stays null, so `refDateTime` is omitted. The device logs the missing reference (`-45`, accepted with warning) and still registers the document.

## Context
Replaces the **rejected #718** (`ToString("o")` emitted a `Z` designator that `DateTime.Parse` converts to host-local, corrupting the CCDC-chained `refDateTime`).

## Tests
Summer (offset 2): 12:00 UTC → device `dateTime="20260702T140000"` + signature `2026-07-02 14:00:00`. Winter (offset 1): 12:00 UTC → 13:00. Conversion verified host-independent for `Utc`/`Unspecified`/`Local` kinds; unreferenced refund omits `refDateTime`. Full EpsonRTServer suite **72/72**.

### DIFF
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerMapping.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerMapping.cs
@@ -35,9 +35,10 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTServer
         {
             var docNumber = tillState.LastDocNumber + 1;
             var zNumber = tillState.LastZNumber;
-            // cbReceiptMoment is UTC, but the RT Server records the LOCAL emission time (Metadata Guide 3.8:
-            // dateTime "YYYYMMDDThhmmss" + srtUtcOffset 1=winter/2=summer). Convert with the device-reported
-            // offset so the time is correct on any host (the cloud host runs in UTC; ToLocalTime would be wrong).
+            // The RT Server records the LOCAL emission time (Metadata Guide 3.8: dateTime "YYYYMMDDThhmmss"
+            // + srtUtcOffset 1=winter/2=summer). cbReceiptMoment is defined as UTC (fiskaltrust interface-doc),
+            // so convert with the device-reported offset — correct on any host (the cloud host runs in UTC,
+            // where ToLocalTime would be wrong).
             var moment = ToRtServerLocalTime(receiptRequest.cbReceiptMoment, tillState.SrtUtcOffset);
 
             var recAmount = GetReceiptTotal(receiptRequest, docType);
@@ -78,13 +79,18 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTServer
         }
 
         /// <summary>
-        /// Converts a UTC receipt moment to the RT Server's local wall-clock time using the device-reported
-        /// <paramref name="srtUtcOffset"/> (1 = winter/+1h, 2 = summer/+2h). Returns an Unspecified-kind value
-        /// so it formats without a timezone designator, as the metadata format requires.
+        /// Converts the receipt moment to the RT Server's local wall-clock time using the device-reported
+        /// <paramref name="srtUtcOffset"/> (1 = winter/+1h, 2 = summer/+2h). cbReceiptMoment is defined as UTC
+        /// (fiskaltrust interface-doc: "Must be provided in UTC"); since the DateTimeKind is not guaranteed
+        /// after deserialization, a Local value is normalised via ToUniversalTime and a zoneless (Unspecified)
+        /// value is honoured as UTC per that contract. Returns an Unspecified-kind value so it formats without
+        /// a timezone designator, as the metadata format requires.
         /// </summary>
         public static DateTime ToRtServerLocalTime(DateTime moment, int srtUtcOffset)
         {
-            var utc = moment.Kind == DateTimeKind.Local ? moment.ToUniversalTime() : moment;
+            var utc = moment.Kind == DateTimeKind.Local
+                ? moment.ToUniversalTime()
+                : DateTime.SpecifyKind(moment, DateTimeKind.Utc);
             return DateTime.SpecifyKind(utc.AddHours(srtUtcOffset), DateTimeKind.Unspecified);
         }
 
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerMapping.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerMapping.cs
@@ -35,7 +35,10 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTServer
         {
             var docNumber = tillState.LastDocNumber + 1;
             var zNumber = tillState.LastZNumber;
-            var moment = receiptRequest.cbReceiptMoment;
+            // cbReceiptMoment is UTC, but the RT Server records the LOCAL emission time (Metadata Guide 3.8:
+            // dateTime "YYYYMMDDThhmmss" + srtUtcOffset 1=winter/2=summer). Convert with the device-reported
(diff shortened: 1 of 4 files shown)


scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerSCU.cs
```diff
@@ -248,10 +248,11 @@ private async Task<List<SignaturItem>> ProcessFiscalDocumentAsync(ReceiptRequest
             }
             else if (string.IsNullOrEmpty(receiptRequest.cbPreviousReceiptReference))
             {
-                // Unreferenced refund/void: use neutral references and the current moment.
+                // Unreferenced refund/void: use neutral references and the current moment (converted to the
+                // RT Server's local time, consistent with the receipt's own dateTime).
                 referenceZNumber = 0;
                 referenceDocNumber = 0;
-                referenceDocMoment = receiptRequest.cbReceiptMoment;
+                referenceDocMoment = EpsonRTServerMapping.ToRtServerLocalTime(receiptRequest.cbReceiptMoment, tillState.SrtUtcOffset);
```
StefanKert: For references we should pull data from the response right? Can you check with the customrtserver

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerMapping.cs
```diff
@@ -74,6 +77,17 @@ public static FiscalDocumentResult BuildFiscalDocument(
             };
         }
 
+        /// <summary>
+        /// Converts a UTC receipt moment to the RT Server's local wall-clock time using the device-reported
+        /// <paramref name="srtUtcOffset"/> (1 = winter/+1h, 2 = summer/+2h). Returns an Unspecified-kind value
+        /// so it formats without a timezone designator, as the metadata format requires.
+        /// </summary>
+        public static DateTime ToRtServerLocalTime(DateTime moment, int srtUtcOffset)
```
StefanKert, line 89: This looks kinda dangerous. So `cbReceiptMoment` is expected to be UTC always ( as per definition) so if we should go with localtime we can do that

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerSCU.cs
accodev: We are building the request to be sent to the server here, how can we pull data from the response?

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerMapping.cs
accodev: how would you proceed to address this issue? I see it's a dangerous assumption, isn't the `cbReceiptMoment` formatted in iso standard so we can infer if the datetime is utc or local time?

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerSCU.cs
StefanKert: As for unreferenced comments ( depending on what the expectation si from a protocol) we also don't have the original receiptmoment.

I am not really sure if this a valid case though and this looks kinda dangerous, but if we don't have the date of the original receipt, we should not set it.

settingthe cbreceiptmoment would use todays date, even though this is very likely wrong

accodev: Referenced refunds/voids already pull from the response: the queue injects the original receipt's `RTReferenceZNumber` / `RTReferenceDocumentNumber` / `RTReferenceDocumentMoment` signatures, and the SCU reads them (242-248) â€” `referenceDocMoment = DateTime.Parse(RTReferenceDocumentMoment)`, i.e. the original document's own (device-local) moment, not `cbReceiptMoment`. Same pattern the CustomRTServer uses in `CreateAnnuloDocument`/`CreateResoDocument`. So for a real reference we already source from the response â€” no change needed there. (7e71e1c)

accodev: Agreed â€” fixed in 7e71e1c. The unreferenced branch no longer sets `cbReceiptMoment` (~today). `referenceDocMoment` is left null so `refDateTime` is omitted entirely; the device then answers `-45` ("reference date omitted"), which is in the accepted-with-warning range, so the document is still registered and the anomaly is logged â€” instead of us inventing a wrong date. Added a test asserting the unreferenced refund emits no `refDateTime`.

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerMapping.cs
accodev: Checked both ways:

- **By contract it *is* UTC** â€” fiskaltrust interface-doc (data-structures): *"The time of receipt creation. Must be provided in UTC, e.g. `2020-06-29T17:45:40.505Z`."*
- **At runtime it is *not* guaranteed** â€” QueuePT explicitly rejects `cbReceiptMoment.Kind != DateTimeKind.Utc` (`ReceiptRequestValidations.cs:160`); that guard only exists because non-UTC values actually arrive, and IT has no equivalent guard.

So rather than assume, `ToRtServerLocalTime` now inspects the `DateTimeKind` (which .NET sets from the ISO-8601 zone designator): `Local` â†’ `ToUniversalTime()`, zoneless (`Unspecified`) â†’ honored as UTC per the contract, then the device-reported `srtUtcOffset` is applied (not `ToLocalTime()`, which would be wrong on the UTC cloud host). Tests now cover all three kinds. (7e71e1c)

StefanKert (approved):

accodev:
/deploy scu-it EpsonRTServer
[](https://github.com/fiskaltrust/middleware/actions/runs/29936840279)

PR #724, opened 2026-07-23
fix(epsonrtserver): realign token on Z-session shift in the offline drain

accodev:
Tracks **fiskaltrust/market-it#631**.

## Problem
Offline (async cache + background drain), when the RT Server crosses a **Z-session boundary** while documents are buffered (forced/daily closure — Metadata Guide §3.4.7/§3.4.5), the SCU kept minting documents for the **closed** session and the drain replayed them → device returned **`-25` "NUMERO DOCUMENTO ERRATO"** on every one. Online is unaffected. (Pre-#720/#722 this surfaced as `-51`; same root cause.)

**Root cause:** the async path never refreshed the token after a session shift — the inline token-reseed recovery only runs on a synchronous response, which async signing never produces.

## Fix
**Part 1 — realign.** On `-21/-22/-25` the drain realigns the till to the device's current session via a fresh token (SCU-owned hook), so subsequent documents build correctly. Realign fires only from the background loop → no lock re-entrancy; daily-closing parks-and-continues instead of throwing.

**Part 2 — consume-if-already-registered.** Live device reads proved these out-of-sync receipts are **already registered** on the device (accepted with error in log — readable by number, returning the prefixed CCDC), *not* rejected. So resending would **double-register**. The drain now reads the doc back by its original coordinates (`createReport/receipt`) and, if the device's `hash` suffix-matches the cached CCDC, **consumes** the redundant local copy + warns; any read failure (`-31`/`-35`/mismatch) **parks** for manual review (never consumed on doubt, never resent).

No renumber/resend, no cache-format change, no new model.

## Tests
`dotnet test` green (**79/79**); builds `netstandard2.0;net461;net6`. New: realign+park, realign reseed, consume-if-registered, park-if-not-registered, daily-closing park-and-continue.

## Known limitation / follow-up (domain)
A device-registered doc flagged `NUMERO DOCUMENTO ERRATO` (with `rtFileRejected=6`): whether it counts as valid AdE takings or is rejected at transmission is a device/AdE-layer question, separate from the SCU send logic. This PR ensures the SCU stops mishandling these and realigns numbering so they stop occurring.

### DIFF
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerClient.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerClient.cs
@@ -93,6 +93,9 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTServer
         public Task<RtServerResponse> GetFiscalInformationAsync(string tillId)
             => SendToFpServerAsync($"<createReport><fiscalInformation tillId=\"{tillId}\" /></createReport>");
 
+        public Task<RtServerResponse> GetReceiptAsync(string tillId, long zRepNumber, long recNumber, string date)
+            => SendToFpServerAsync($"<createReport><receipt tillId=\"{tillId}\" zRepNumber=\"{zRepNumber:D4}\" recNumber=\"{recNumber:D4}\" date=\"{date}\" reportType=\"txt\" /></createReport>");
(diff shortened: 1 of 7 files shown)


accodev:
/deploy scu-it EpsonRTServer
[](https://github.com/fiskaltrust/middleware/actions/runs/30003085472)

danielscainifiskaltrust (approved):

PR #727, opened 2026-07-23
fix(epsonrtserver): map VAT from VATRate when ftChargeItemCase is 0x0 (market-it #634)

accodev:
## Problem
A receipt whose charge items carry **no `ftChargeItemCase`** (low nibble `0x0`) but a valid `VATRate` is rejected by the Epson RT Server SCU with `NotSupportedException: The ftChargeItemCase 0x0 has no VAT-index mapping`, failing the whole receipt. Reproduced with a plain cash sale (`VATRate: 22`, no `ftChargeItemCase`). (Originally reported as a "copy receipt" crash; the RT Server has no copy command — the real trigger is the missing case nibble.)

## Fix (minimal)
`TryGetVatId`: add a `case 0x0` that maps the item's explicit `VATRate` to the RT Server VAT index for the standard Italian rates (22→1, 10→2, 5→4, 4→3). `default` is unchanged, so genuinely unknown goods (unrecognised nibble, or `0x0` with a non-standard/zero rate) still throw — the "wrong rate is worse than failing" invariant is preserved. Consistent with the sibling CustomRTServer, which already trusts `VATRate` when set.

## Tests
Added a Theory asserting the `0x0`+`VATRate` fallback for each standard rate, and a Fact asserting `0x0` with an unrecognised rate still throws `NotSupportedException`. RED first (fallback threw), GREEN after. Full EpsonRTServer unit suite 84/84.

Refs: fiskaltrust/market-it#634 (the follow-on host 500 is fixed separately in the #635 PR).

### DIFF
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerMapping.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTServer/EpsonRTServerMapping.cs
@@ -418,8 +418,21 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTServer
                 case 0x3: vatId = 1; return true;  // 22%
                 case 0x4: vatId = 4; return true;  // 5%
                 case 0x7: vatId = 13; return true; // 0%
+                // No VAT nibble (0x0): honour the item's stated VATRate so receipts that omit ftChargeItemCase
+                // are not rejected; unrecognised rates stay strict (fall through to throw).
+                case 0x0: return TryGetVatIdFromVatRate(chargeItem.VATRate, out vatId);
                 default: vatId = -1; return false;
             }
+        }
+
+        private static bool TryGetVatIdFromVatRate(decimal vatRate, out int vatId)
+        {
+            if (vatRate == 22m) { vatId = 1; return true; }
+            if (vatRate == 10m) { vatId = 2; return true; }
+            if (vatRate == 5m) { vatId = 4; return true; }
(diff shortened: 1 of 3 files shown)


danielscainifiskaltrust (dismissed):

accodev:
Device-confirmed on the Fiberland RT Server (serial `99SEA004010`, till `FISK0001`).

Sent a POS receipt with four items, each omitting the VAT nibble (`ftChargeItemCase = 0x4954_2000_0000_0000`) and carrying only a numeric `VATRate`, one per active rate. The `<createReport><fiscalInformation>` read-back exposes the device's own VAT table, which matches this mapping exactly:

| emitted `vatID` | device `vatCount` rate | item gross | booked into bucket |
|---|---|---|---|
| 1 | 22,00% | 12.20 | +12.20 |
| 2 | 10,00% | 11.00 | +11.00 |
| 3 | 4,00%  | 10.40 | +10.40 |
| 4 | 5,00%  | 10.50 | +10.50 |

Receipt accepted cleanly (`code=0 OK`, `POSRECEIPT`, no `rt-server-receipt-warning`), and each item landed in the correct rate bucket (per-rate delta == item gross across two consecutive runs). This resolves the open question of whether the assumed index order (`22->1, 10->2, 4->3, 5->4`) matches the device programming — it does.

Added as an acceptance test: `EpsonRTServerTests.ProcessPosReceipt_VatFallbackFromVatRate_ShouldBeAcceptedCleanly`.

danielscainifiskaltrust (approved):

PR #633, opened 2026-03-09
#306 - Update of Swissbit Hardware TSE V2

forsthug:
- [ ] You've read the [Contributor Guide](https://github.com/fiskaltrust/.github/blob/main/CONTRIBUTING.md) and [Code of Conduct](https://github.com/fiskaltrust/.github/blob/main/CODE_OF_CONDUCT.md).
- [ ] You've included unit or integration tests for your change, where applicable.
- [ ] You've included inline docs for your change, where applicable.
- [ ] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

{Summary of the changes}

## Description

{Detail}

Fixes #{issue number}

PR #729, opened 2026-07-23
fix(it): handle unreferenced refund/void — ND header + refRefundVoidType (market-it #633)

accodev:
Two independent parts of the unreferenced-refund flow.

## Part 1 — QueueIT crash (the reported bug)
`SignaturItemFactory.CreateHeader` did `DateTime.Parse(referenceDateTimeString)` on the REFUND/VOID "ND" branch. For an **unreferenced** refund/void the RT reference date is absent (the SCU omits it), so `referenceDateTimeString` is null → `ArgumentNullException`, failing the receipt at the queue **even though the device registered the document**. Fix: render `"ND"` (no date) when the reference date is empty. Shared across all IT SCUs; additive guard (every currently-working path has a non-null date and is unchanged).

## Part 2 — device `-35` (refRefundVoidType="ND")
For the unreferenced case the SCU emitted zeroed `refZRepNum="0000"`/`refRecNum="0000"`, which the RT Server logs as *"refund/void with wrong date reference"* (**-35**). Per Metadata Guide §3.7.2, an untraceable refund/void is type **ND** (for POS/VR/ND only `docType` + `refRefundVoidType` apply, not reference numbers). The unreferenced `beginFiscalReceipt` now emits `refRefundVoidType="ND"` instead of the zeroed references. Referenced refunds/voids are unchanged.

## Tests
- QueueIT: unreferenced refund renders `ND` without throwing (RED→GREEN). 16/16.
- EpsonRTServer: unreferenced refund emits `refRefundVoidType="ND"` and no zeroed refs (RED→GREEN); referenced-refund test unchanged. 80/80.

## Follow-up
Part 2 changes device XML — confirm on the Fiberland device that an unreferenced refund now closes without `-35` (and whether the device additionally wants a `refDateTime` for ND; omitted here to avoid inventing a reference date).

Refs: fiskaltrust/market-it#633

### DIFF
--- a/queue/src/fiskaltrust.Middleware.Localization.QueueIT/Factories/SignaturItemFactory.cs
+++ b/queue/src/fiskaltrust.Middleware.Localization.QueueIT/Factories/SignaturItemFactory.cs
@@ -106,7 +106,9 @@ namespace fiskaltrust.Middleware.Localization.QueueIT.Factories
                 stringBuilder.AppendLine("emesso per RESO MERCE");
                 if (string.IsNullOrEmpty(referenceZNumberString) || string.IsNullOrEmpty(referenceDocNumberString))
                 {
-                    stringBuilder.AppendLine($"ND del {DateTime.Parse(referenceDateTimeString).ToString("dd-MM-yyyy")}");
+                    stringBuilder.AppendLine(string.IsNullOrEmpty(referenceDateTimeString)
+                        ? "ND"
+                        : $"ND del {DateTime.Parse(referenceDateTimeString).ToString("dd-MM-yyyy")}");
                 }
                 else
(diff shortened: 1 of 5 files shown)


accodev:
✅ **Fiberland device confirmation.** First run (refRefundVoidType="ND" only): -35 cleared, but the device then warned **-45** ("Refund or void reference document error (time omitted or zero)") — the RT Server requires a refDateTime for ND. Follow-up commit emits the receipt's own moment as refDateTime; the unreferenced refund is now accepted **with no warning at all** (doc 0003, REFUND). Added a device acceptance test.

danielscainifiskaltrust (approved):

PR #748, opened 2026-08-05
[IT] EpsonRTPrinter: fall back to cached serial number in documento commerciale signatures

### LINKED ISSUE #743: [IT] EpsonRTPrinter: reprint (0x3010) emits <rt-serialnumber>signature withoutData`, causing a 400 deserialization error downstream
When processing a copy/reprint receipt (ftReceiptCase 0x...3010) with the EpsonRTPrinter SCU, the middleware returns a receipt response whose first signature item (<rt-serialnumber>) has no Data property. The PosSystem API (fiskaltrust.Api.PosSystem.Web → MiddlewareClient.SignAsync) deserializes the response with Data marked as required, so the whole request fails with HTTP 400 — even though the reprint was actually performed on the printer and the receipt was persisted in the queue.

Newtonsoft.Json.JsonSerializationException: Required property 'Data' not found in JSON. Path 'ftSignatures[0]'
   at ...
   at fiskaltrust.Api.PosSystem.Web.Signing.MiddlewareClient.SignAsync(...) in MiddlewareClient.cs:line 85

Response fragment produced by the middleware (note the first signature):

"ftSignatures":[
  {"ftSignatureFormat":1,"ftSignatureType":...3280,"Caption":"<rt-serialnumber>"},
  {"ftSignatureFormat":1,"ftSignatureType":...3281,"Caption":"<rt-z-number>","Data":"0000"},
  {"ftSignatureFormat":1,"ftSignatureType":...3282,"Caption":"<rt-doc-number>","Data":"0000"},
  ...
]

Root cause

1. EpsonRTPrinterSCU.ProcessPerformReprint executes the reprint via directIO command="3098" (EpsonCommandFactory.ReprintCommand). The Epson response to a directIO command does not contain the receiptInfo block — no serial number, no fiscal receipt number, no Z-report number, no date/time.
2. The serial number is then taken from that response without any fallback (EpsonRTPrinterSCU.cs:579):
RTSerialNumber = result?.Receipt?.SerialNumber,   // → null
2. Other code paths in the same file use ?? "" (lines 312, 383, 436) or the cached _serialnr. The same unguarded pattern also exists at lines 661 and 735 (refund/void paths).
3. SignatureFactory.CreateDocumentoCommercialeSignatures (SignatureFactory.cs:75) assigns the null to SignaturItem.Data; JSON serialization omits the null property, producing a signature item without Data, which the consumer contract treats as required.

Supporting evidence in the same response, all consequences of the empty directIO receiptInfo:
- rt-z-number / rt-doc-number are 0000 (SetReceiptResponse falls back to 0, EpsonRTPrinterSCU.cs:156-157),
- rt-doc-moment is DateTime.Now (EpsonRTPrinterSCU.cs:166, already flagged with a // ??????? comment).

Impact

- Any reprint through the EpsonRTPrinter SCU fails with HTTP 400 at the PosSystem API layer.
- The failure happens after the printer has printed and the queue has persisted the receipt (the 400 response even carries a Location: /v2/PeekJournalItem/... header), so a client retry produces a duplicate physical reprint.
- The refund/void paths (lines 661, 735) share the same latent bug.

Proposed fix

1. Primary — in EpsonRTPrinterSCU, fall back to the cached serial number, as CustomRTPrinterSCU already does (CustomRTPrinterSCU.cs:704 uses _serialnr):
RTSerialNumber = result?.Receipt?.SerialNumber ?? _serialnr ?? "",
1. at lines 579, 661 and 735.
2. Defensive — in SignatureFactory.CreateDocumentoCommercialeSignatures, guard Data = data.RTSerialNumber ?? "" so no code path can ever emit a signature item without Data.

Reproduction

1. Configure an EpsonRTPrinter SCU (queue IT).
2. Sign a regular receipt, then send a copy receipt (ftReceiptCase 0x4954200000003010) referencing it via cbPreviousReceiptReference.
3. Observe the 400 response from the PosSystem API with Required property 'Data' not found in JSON. Path 'ftSignatures[0]', while the printer performs the reprint.

danielscainifiskaltrust:
Fixes #743.

### Problem

Reprints (`0x3010`) on the EpsonRTPrinter SCU run through directIO `3098`. That response carries no `receiptInfo` block, so `ProcessPerformReprint` built the `<rt-serialnumber>` signature with `Data = null`. JSON serialization omits a null property, and consumers that require the field fail to deserialize the response with a 400, even though the printer completed the reprint. The refund and void paths carried the same latent issue.

The 400 arrives after the printer printed and after the queue stored the receipt, so a client retry prints a second copy.

Confirmed again on a device on 2026-08-06. The response held `<rt-serialnumber>` without `Data`, `rt-z-number` `0000`, `rt-doc-number` `0000`, `rt-doc-moment` at the request time, and `ftState` `0x4954200000000000`, the plain success state.

### Fix

- `ProcessPerformReprint` reads the serial number through `GetRTInfoAsync()` when the cached value is empty, and it reads it **before** it sends the reprint command. `GetRTInfoAsync` sends two commands to the printer and it can throw. A failure before the print errors the receipt with no document on paper. A failure after the print would error a document that already exists. The refund path (line 627) and the void path (line 710) already use this order.
- All `POSReceiptSignatureData` construction sites in `EpsonRTPrinterSCU` use `result?.Receipt?.SerialNumber ?? _serialnr ?? ""` instead of a value that can be null.
- Last-resort guard in `SignatureFactory.CreateDocumentoCommercialeSignatures`: `Data = data.RTSerialNumber ?? ""`.

### Scope check (other IT SCUs)

`POSReceiptSignatureData.RTSerialNumber` is declared `string`, not `string?`. Nullable is enabled, and `TreatWarningsAsErrors` is commented out in `scu-it/Directory.Build.props:13`, so the compiler warnings for these assignments go unread. Eight sites outside the reprint path assign a value that can be null:

| SCU | Sites | Source |
| --- | --- | --- |
| CustomRTPrinter | 300, 421, 518, 659, 704 | `_serialnr`, declared `string?` |
| CustomRTServer | 333 | `cashuuid.RTServerSerialNumber` |
| EpsonRTServer | 306 | `tillState.RTServerSerialNumber` |

None of the three SCUs handles `0x3010`, so none reproduces #743. Each one can still emit a null serial number on its own paths when the serial is unknown. The new guard in `SignatureFactory` covers all of them. A change of those eight sites belongs in a separate PR.

### Tests

- New `ReprintSerialNumberTests` covering `ProcessPerformReprint` with a mocked `IEpsonFpMateClient`:
  - a `3098` response without an `addInfo` block signs the serial number of the printer,
  - a printer that reports no `rtType` signs `""` and never `null`,
  - the SCU reads the serial number before it sends the reprint command.
- All three tests fail against `main` and pass with this change.
- New `SignatureFactoryTests` covering the null-serial guard.

Local results: EpsonRTPrinter 12/12, SCU.IT UnitTest 5/5, CustomRTServer 32 passed with 4 skipped, EpsonRTServer 85/85. The CustomRTPrinter suite stays device-gated and reports 32 skipped.

### Open point

Device validation of the fix on a physical Epson RT printer is still pending.

The reprint signature still reports `rt-z-number` `0000`, `rt-doc-number` `0000` and `rt-doc-moment` at the request time, because the `3098` response carries none of those values. The original document values remain available in `rt-reference-z-number`, `rt-reference-doc-number` and `rt-reference-doc-moment`. A change of the three fields is a behaviour change beyond #743 and needs its own issue.

### DIFF
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.Abstraction/SignatureFactory.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.Abstraction/SignatureFactory.cs
@@ -72,8 +72,8 @@ public static class SignatureFactory
             new SignaturItem
             {
                 Caption = "<rt-serialnumber>",
-                // Data is required by consumers deserializing the receipt response; a null here
-                // gets omitted from the serialized JSON and breaks their contract.
+                // Last resort against a null from any SCU: consumers require Data, and JSON
+                // serialization omits the property when it is null (#743).
                 Data = data.RTSerialNumber ?? "",
                 ftSignatureFormat = (long) SignaturItem.Formats.Text,
                 ftSignatureType = ITConstants.BASE_STATE |(long) SignatureTypesIT.RTSerialNumber
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.Abstraction/SignatureFactory.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.Abstraction/SignatureFactory.cs
@@ -72,7 +72,9 @@ public static class SignatureFactory
             new SignaturItem
             {
                 Caption = "<rt-serialnumber>",
-                Data = data.RTSerialNumber,
+                // Data is required by consumers deserializing the receipt response; a null here
+                // gets omitted from the serialized JSON and breaks their contract.
+                Data = data.RTSerialNumber ?? "",
                 ftSignatureFormat = (long) SignaturItem.Formats.Text,
                 ftSignatureType = ITConstants.BASE_STATE |(long) SignatureTypesIT.RTSerialNumber
             },
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/EpsonRTPrinterSCU.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/EpsonRTPrinterSCU.cs
@@ -253,7 +253,7 @@ public sealed class EpsonRTPrinterSCU : LegacySCU
             }
             var posReceiptSignatur = new POSReceiptSignatureData
             {
-                RTSerialNumber = result?.Receipt?.SerialNumber,
+                RTSerialNumber = result?.Receipt?.SerialNumber ?? _serialnr ?? "",
                 RTZNumber = fiscalReceiptResponse.ZRepNumber,
                 RTDocNumber = fiscalReceiptResponse.ReceiptNumber,
                 RTDocMoment = fiscalReceiptResponse.ReceiptDateTime,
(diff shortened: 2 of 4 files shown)


accodev:
/deploy scu-it EpsonRTPrinter
[](https://github.com/fiskaltrust/middleware/actions/runs/31079359999)

accodev:
/deploy scu-it EpsonRTPrinter
[](https://github.com/fiskaltrust/middleware/actions/runs/31084574604)

accodev:
Confirmed working on real hardware.

mttgennarift (approved):

PR #756, opened 2026-08-27
Update actions

LeilaZarban:

{Summary of the changes}

## Description

{Detail}

Fixes #{issue number}


LeilaZarban:
/package queue SQLite
[](https://github.com/fiskaltrust/middleware/actions/runs/33086664778)

LeilaZarban:
/deploy queue SQLite
[](https://github.com/fiskaltrust/middleware/actions/runs/33088166302)

.github/actions/version/action.yml
```diff
@@ -0,0 +1,27 @@
...
+outputs:
+  version:
+    description: "The version of the package"
+    value: ${{ steps.get-version.outputs.version }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install Nerdbank.GitVersioning
+      run: dotnet tool install -g nbgv
```
pawelvds, line 17: Checks are green here, but did you run into any errors with this? I had a case in `middleware-launcher` where `nbgv` was already installed on the runner (3.10.94 on `ubuntu-latest`), so the regular `install` failed. Just wanted to mention it in case the same issue comes up here.

.github/workflows/deploy.yml
```diff
@@ -59,7 +59,7 @@ jobs:
         uses: fiskaltrust/middleware/.github/actions/deploy-v2-packages@main
         with:
           directory: artifacts/package-v2-${{ inputs.package }}
-          package: ${{ inputs.package }}
+          destinationPath: undefined/${{ inputs.package }}
```
pawelvds, line 62: Is `undefined/` here on purpose? It looks like it used to be hardcoded inside the old action, and now it's just copied into the caller. Our `deploy.yml` passes a real path instead (e.g. `win-x64/fiskaltrust.Launcher/`), no `undefined/` prefix.

pawelvds (approved):

.github/workflows/deploy.yml
LeilaZarban: middleware packages are deployed to undefined. it's correct

.github/actions/version/action.yml
LeilaZarban: In middleware, the version action runs only on Windows. 
If we have a problem with middleware-launcher, we should find a solution like run it on windows and store it the version in github variables and distribute it to other jobs.

PR #760, opened 2026-09-01
feat(scu-gr): support myDATA 11.3 simplified invoice

AVPavel:
## Problem

Adds first-class support for myDATA document type **11.3 – Απλοποιημένο Τιμολόγιο
(Simplified Invoice)**. 11.3 is an invoice to a *business* in legally simplified
form (ΕΛΠ ν.4308/2014 άρθρο 10): the buyer's details are omitted, so myDATA files
it as non-counterparty, but because the buyer is a business, the revenue is
classified as **wholesale**, not retail.

Previously the type could only be reached through a `mydataoverride`, and even then
the income classification stayed on the retail branch (`E3_561_003`), producing a
document AADE accepts but books incorrectly.

## Solution

- **New GR-local flag** `IsSimplifiedInvoice = 0x0000_0001_0000_0000` (lll region),
  added to both `ReceiptCaseFlagsGR` copies.
- **`GetInvoiceType`** resolves a POS receipt carrying the flag to `Item113`,
  evaluated *before* the goods/services split so it applies to both. Refund still
  wins (→ `11.4`), as do agency (`11.5`) and own-consumption (`6.1`).
- **Wholesale income classification.** The base mapping stays a pure function of the
  receipt case, it produces the retail `E3_561_003`, as for 11.1. A single
  normalization step in `AADEFactory`, applied *after* the override layer, remaps it
  to the wholesale `E3_561_001` when the resolved document is a simplified invoice:
  recognized by the invoice type being `11.3` (covers the override route) or by the
  `IsSimplifiedInvoice` flag (covers a refund that resolves to `11.4`). Only the
  retail code is remapped, so an explicit per-line `incomeClassification` override is
  preserved. This keeps the override out of the case-driven base mapper (matching the
  existing `GetInvoiceType` design) and also fixes the pre-existing override path,
  which had been transmitting the retail code.
- **No counterpart:** `SupportsCounterpart(Item113)` / `RequiresCustomerInfo(Item113)`
  stay `false`, a supplied `cbCustomer` is not transmitted, matching the
  non-counterparty nature of the type.
- **Tests:** `SimplifiedInvoice113Tests` covers the trigger (goods + services),
  precedence (refund/agency/own-consumption), the wholesale classification asserted on
  the built document for both the flag and the override routes (and that plain retail
  and plain refunds stay retail), the dropped customer, and correlation via
  `multipleConnectedMarks`.

## Behaviour notes

- **Thresholds** (€100 general, €300 fuel, none for corrective documents) are the
  POS's legal responsibility and are not enforced by the middleware.
- **Reverse charge (άρθρο 39a) on 11.3** is not a supported combination and is left
  unmapped for now (documented inline).
- **Viva:** no gateway change needed

Fixes: [#265](https://github.com/fiskaltrust/market-gr/issues/265)

### DIFF
--- a/queue/src/fiskaltrust.Middleware.Localization.QueueGR/Models/Cases/ReceiptCaseFlagsGR.cs
+++ b/queue/src/fiskaltrust.Middleware.Localization.QueueGR/Models/Cases/ReceiptCaseFlagsGR.cs
@@ -4,7 +4,16 @@ namespace fiskaltrust.Middleware.Localization.QueueGR.Models.Cases;
 
 public enum ReceiptCaseFlags : long
 {
-    HasTransportInformation = 0x0000_0000_0400_0000
+    HasTransportInformation = 0x0000_0000_0400_0000,
+
+    /// <summary>
+    /// local flag for:
+    /// Απλοποιημένο Τιμολόγιο — simplified invoice, myDATA document type 11.3.
+    /// An INVOICE to a business in legally simplified form (ΕΛΠ ν.4308/2014 άρθρο 10), not a retail
+    /// receipt: the buyer's details are omitted, which is why 11.3 is a
+    /// «Μη Αντικριζόμενο» / non-counterparty document.
+    /// </summary>
+    IsSimplifiedInvoice = 0x0000_0001_0000_0000
 }
 
 public static class ReceiptCaseFlagsExt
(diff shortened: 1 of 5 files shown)


StefanKert:
Closing, we'll use an override for now

PR #769, opened 2026-09-08
fix(scu-it/epson): redeem a multi-use voucher charge item as sconto a pagare

### LINKED ISSUE #514: Multi-Use Voucher redeem as chargeitem leads to error with Epson RT PRinter

accodev:
Fixes #514

## Why

A multi-use voucher (buono multiuso) can be redeemed in two spellings of the fiskaltrust API: as pay item `0x06`, or as a negative charge item `0x…0048` (type of service 4, VAT nibble 8). The Epson RT Printer SCU handled only the first. `EpsonCommandFactory.GenerateItems` mapped *every* `0x…48` line to a `printRecItem` on department 11 with the **signed** quantity and unit price, so a redeem produced

```xml
<printRecItem description="Gutschein" quantity="1" unitPrice="-20" department="11" />
```

Item amounts carry no sign on the Epson protocol, and the printer refuses the command (`success="false" code="PRINTER ERROR" status="16"`, `lastCommand` 80), so the whole receipt failed. In grouping requests the same line became a `printRecVoidItem` (a void of an item that was never sold), and in refund/void documents it went out on department `-1`.

On the RT a multi-use voucher redemption is neither an item nor money received: the Agenzia delle Entrate XML puts it into `ScontoApagare` ("va indicato anche l'importo dei pagamenti effettuati con buono multiuso", gross of VAT, no effect on IVA). The Epson protocol 7.00 (§9.13) implements this as payment **type 6, index 01** ("Increments Sconto a pagare. Does not increment Importo pagato"), which is exactly what pay item `0x06` already produces.

## What changes

- **`ReceiptCaseHelper.IsMultiUseVoucherRedeem(chargeItem, request)`** (shared Abstraction): a `0x…48` line with a negative amount on a sale, or a positive amount on a refund/void (which mirror the sale). A positive line on a sale stays what it is: the sale of the voucher itself.
- **`EpsonCommandFactory.GetTotalAndMessages`** emits one `printRecTotal paymentType="6" index="1"` for `|Amount|` per redeem line, **before** the pay items (protocol note [2]: a type 6 payment above the amount still due is refused with error 21). The method is shared by the invoice, refund and void builders, so one change covers all three documents. A voucher that covers the whole receipt completes the payment phase on its own; the zero-cash fallback no longer fires.
- **`GetItemAndMessages`, `GetRecRefunds`, `GetRecvoids`** skip the redeem line. A refund/void of a voucher **sale** line now uses the NS department 11 like the sale (was `GetVatGroup()` = `-1`).
- **Tests:** `MultiUseVoucherRedeemTests` (10 cases: sale, grouping request, voucher covers the whole receipt, refund, void, and two regression guards for the voucher sale line). New acceptance example `0x0001_CashWithMultiUseVoucherRedeem_AsChargeItem.json` with helper and fact for the hardware suite.

## Validation on hardware

Sandbox Epson RT `99IEB077964` (training mode), driven with the XML the SCU generates:

| Step | Document | Printer response |
|---|---|---|
| Pre-fix XML (item with `unitPrice="-20"` on dept 11) | Z 98, doc 2 aborted | `PRINTER ERROR`, status 16 |
| Sale: Food 100 @10 %, voucher −20 as charge item, cash 80 | Z 98, doc 1 | accepted, amount 100,00 |
| Refund referencing doc 1 (voucher as type 6, cash 80) | Z 98, doc 3 | accepted |
| Second sale, then void referencing it | Z 98, docs 4 and 5 | accepted |
| Sale paid 100 % by the voucher (no pay items) | Z 98, doc 6 | accepted |
| Sale: Food 100, voucher −20, **no pay items** → voucher 20 + cash `payment="0"` (review round 2) | Z 98, doc 7 | accepted, amount 100,00 |

Unit tests: all 58 Epson unit tests pass; the other scu-it unit test projects stay green; the scu-it solution builds.

## Release note

Epson RT Printer: a multi-use voucher redeemed as a negative charge item (`…0048`) is now sent to the printer as a "sconto a pagare" payment (type 6, index 1) in sale, refund and void documents, instead of a negative item line the printer rejected. Refunds and voids of a multi-use voucher sale line now use the NS department 11 like the sale.

## Out of scope (separate tickets)

- **Sibling SCUs → #771.** `EpsonRTServerMapping` and `CustomRTPrinterSCU` print a negative `0x…48` line as a *positive* NS sale line (`Math.Abs`), so the total is wrong there instead of erroring; `CustomRTPrinterSCU.GetPaymentType` maps pay item `0x06` to cash; `CustomRTServerMapping` has no `0x…48` rule at all. (The `scu-it-epsonrtserver` / `scu-it-customrt*` labels on this PR come from `.github/labeler.yml`, which maps the shared Abstraction folder to all four SCU labels.)
- **Tips in refund/void** also land on department `-1`; not touched here.
- **Refund requests with non-mirrored signs** (positive amounts under the refund flag) still take the old path.

### DIFF
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
@@ -472,7 +472,7 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Utilities
                 Quantity = Math.Abs(p.Quantity),
                 UnitPrice = p.Quantity == 0 || p.Amount == 0 ? 0 : Math.Abs(p.Amount) / Math.Abs(p.Quantity),
                 Amount = Math.Abs(p.Amount),
-                Department = p.IsMultiUseVoucher() ? 11 : p.GetVatGroup()
+                Department = p.GetVatGroup()
             }).ToList();
         }
 
@@ -484,7 +484,7 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Utilities
                 Quantity = Math.Abs(p.Quantity),
                 UnitPrice = p.Quantity == 0 || p.Amount == 0 ? 0 : Math.Abs(p.Amount) / Math.Abs(p.Quantity),
                 Amount = Math.Abs(p.Amount),
-                Department = p.IsMultiUseVoucher() ? 11 : p.GetVatGroup()
+                Department = p.GetVatGroup()
             }).ToList();
         }
 
@@ -641,7 +641,7 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Utilities
                         Description = i.Description,
                         Quantity = i.Quantity,
                         UnitPrice = i.Quantity == 0 || i.Amount == 0 ? 0 : i.Amount / i.Quantity,
-                        Department = 11,
+                        Department = _departmentNS,
                     };
                     itemAndMessages.Add(new() { PrintRecItem = printRecItem });
                 }
@@ -655,17 +655,6 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Utilities
                         Department = i.GetVatGroup(),
                     };
                     itemAndMessages.Add(new() { PrintRecItemAdjustment = printRecItemAdjustment });
-                }
-                else if (i.IsMultiUseVoucher())
-                {
-                    var printRecItem = new PrintRecItem
-                    {
-                        Description = i.Description,
-                        Quantity = i.Quantity,
-                        UnitPrice = i.Quantity == 0 || i.Amount == 0 ? 0 : i.Amount / i.Quantity,
-                        Department = 11,
-                    };
-                    itemAndMessages.Add(new() { PrintRecItem = printRecItem });
                 }
                 else if (i.Amount < 0)
                 {
@@ -725,9 +714,8 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Utilities
         {
             var totalAndMessages = new List<TotalAndMessage>();
 
-            // A multi-use voucher redeemed as a charge item is paid like pay item 0x06: "sconto a pagare" for
-            // buoni multiuso (payment type 6, index 1; Epson protocol 7.00 §9.13). It goes before the other
-            // payments because a type 6 payment above the amount still due is rejected with error 21.
+            // A multi-use voucher redeemed as a charge item is paid like pay item 0x06. It goes before the other
+            // payments because a "sconto a pagare" above the amount still due is rejected with error 21.
             foreach (var voucher in request.cbChargeItems.Where(x => x.IsMultiUseVoucherRedeem(request)))
             {
                 totalAndMessages.Add(new()
@@ -735,8 +723,8 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Utilities
                     PrintRecTotal = new PrintRecTotal
                     {
                         Description = voucher.Description,
-                        PaymentType = 6,
-                        Index = 1,
+                        PaymentType = _multiUseVoucherPayment.PaymentType,
+                        Index = _multiUseVoucherPayment.Index,
                         Payment = Math.Abs(voucher.Amount)
                     }
                 });
@@ -781,6 +769,9 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Utilities
             public int Index;
         }
 
+        /// <summary>"Sconto a pagare" for buoni multiuso (Epson protocol 7.00 §9.13): pay item 0x06 and a redeemed multi-use voucher charge item.</summary>
+        private static readonly EpsonPaymentType _multiUseVoucherPayment = new() { PaymentType = 6, Index = 1 };
+
         public static EpsonPaymentType GetEpsonPaymentType(PayItem payItem)
         {
             return (payItem.ftPayItemCase & 0xFF) switch
@@ -791,7 +782,7 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Utilities
                 0x03 => new EpsonPaymentType() { PaymentType = 1, Index = 0 },
                 0x04 => new EpsonPaymentType() { PaymentType = 2, Index = 1 },
                 0x05 => new EpsonPaymentType() { PaymentType = 2, Index = 1 },
-                0x06 => new EpsonPaymentType() { PaymentType = 6, Index = 1 },
+                0x06 => _multiUseVoucherPayment,
                 0x07 => new EpsonPaymentType() { PaymentType = 5, Index = 0 },
                 0x08 => new EpsonPaymentType() { PaymentType = 5, Index = 0 },
                 0x09 => new EpsonPaymentType() { PaymentType = 5, Index = 3 },
@@ -816,18 +807,32 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Utilities
         private static int _vatRateSuperReduced2;
         private static int _vatRateParking;
 
+        // Departments pre-programmed with the non-VAT natures (see GetVatInfo for the printed captions).
+        private static readonly int _departmentEE = 10;
+        private static readonly int _departmentNS = 11;
+        private static readonly int _departmentNI = 12;
+        private static readonly int _departmentES = 13;
+        private static readonly int _departmentRM = 14;
+        private static readonly int _departmentAL = 15;
+
         public static int GetVatGroup(this ChargeItem chargeItem)
         {
+            if (chargeItem.IsMultiUseVoucher())
+            {
+                // A buono multiuso is outside the VAT scope (natura N2 "non soggetta"), whatever the nature bits say.
+                return _departmentNS;
+            }
+
             if ((chargeItem.ftChargeItemCase & 0xF) == 0x8)
             {
                 return (chargeItem.ftChargeItemCase & 0xF000) switch
                 {
-                    0x8000 => 10,
-                    0x2000 => 11,
-                    0x1000 => 12,
-                    0x3000 => 13,
-                    0x4000 => 14,
-                    0x5000 => 15,
+                    0x8000 => _departmentEE,
+                    0x2000 => _departmentNS,
+                    0x1000 => _departmentNI,
+                    0x3000 => _departmentES,
+                    0x4000 => _departmentRM,
+                    0x5000 => _departmentAL,
                     _ => _vatRateUnknown // ?
                 };
             }
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
@@ -465,34 +465,37 @@ namespace fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Utilities
 
         public static List<PrintRecRefund> GetRecRefunds(ReceiptRequest receiptRequest)
         {
-            return receiptRequest.cbChargeItems?.Select(p => new PrintRecRefund
+            // A redeemed multi-use voucher is a "sconto a pagare" payment, not a line item (see GetTotalAndMessages).
+            return receiptRequest.cbChargeItems?.Where(p => !p.IsMultiUseVoucherRedeem(receiptRequest)).Select(p => new PrintRecRefund
             {
                 Description = p.Description,
                 Quantity = Math.Abs(p.Quantity),
                 UnitPrice = p.Quantity == 0 || p.Amount == 0 ? 0 : Math.Abs(p.Amount) / Math.Abs(p.Quantity),
                 Amount = Math.Abs(p.Amount),
-                Department = p.GetVatGroup()
+                Department = p.IsMultiUseVoucher() ? 11 : p.GetVatGroup()
             }).ToList();
         }
 
         public static List<PrintRecVoid> GetRecvoids(ReceiptRequest receiptRequest)
         {
-            return receiptRequest.cbChargeItems?.Select(p => new PrintRecVoid
+            return receiptRequest.cbChargeItems?.Where(p => !p.IsMultiUseVoucherRedeem(receiptRequest)).Select(p => new PrintRecVoid
             {
                 Description = p.Description,
                 Quantity = Math.Abs(p.Quantity),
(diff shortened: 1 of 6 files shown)


scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
```diff
@@ -465,34 +465,37 @@ public static FiscalReceipt CreateVoidRequestContent(EpsonRTPrinterSCUConfigurat
...
-            return receiptRequest.cbChargeItems?.Select(p => new PrintRecRefund
+            // A redeemed multi-use voucher is a "sconto a pagare" payment, not a line item (see GetTotalAndMessages).
+            return receiptRequest.cbChargeItems?.Where(p => !p.IsMultiUseVoucherRedeem(receiptRequest)).Select(p => new PrintRecRefund
             {
                 Description = p.Description,
                 Quantity = Math.Abs(p.Quantity),
                 UnitPrice = p.Quantity == 0 || p.Amount == 0 ? 0 : Math.Abs(p.Amount) / Math.Abs(p.Quantity),
                 Amount = Math.Abs(p.Amount),
-                Department = p.GetVatGroup()
+                Department = p.IsMultiUseVoucher() ? 11 : p.GetVatGroup()
```
accodev: The department should not be hardcoded. Find a cleaner solution

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
```diff
@@ -465,34 +465,37 @@ public static FiscalReceipt CreateVoidRequestContent(EpsonRTPrinterSCUConfigurat
...
         {
-            return receiptRequest.cbChargeItems?.Select(p => new PrintRecVoid
+            return receiptRequest.cbChargeItems?.Where(p => !p.IsMultiUseVoucherRedeem(receiptRequest)).Select(p => new PrintRecVoid
             {
                 Description = p.Description,
                 Quantity = Math.Abs(p.Quantity),
                 UnitPrice = p.Quantity == 0 || p.Amount == 0 ? 0 : Math.Abs(p.Amount) / Math.Abs(p.Quantity),
                 Amount = Math.Abs(p.Amount),
-                Department = p.GetVatGroup()
+                Department = p.IsMultiUseVoucher() ? 11 : p.GetVatGroup()
```
accodev: Same here, do not hardcode

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
```diff
@@ -721,6 +724,24 @@ public static List<PrintRecSubtotalAdjustment> GetSubtotalAdjustments(ReceiptReq
...
+                    PrintRecTotal = new PrintRecTotal
+                    {
+                        Description = voucher.Description,
+                        PaymentType = 6,
+                        Index = 1,
+                        Payment = Math.Abs(voucher.Amount)
+                    }
+                });
+            }
+
```
accodev: Do not hardcode values.

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
accodev: Done in 24240369: `GetVatGroup()` now returns the NS department for a multi-use voucher itself (a buono multiuso is natura N2 whatever the nature bits say), so this is back to plain `p.GetVatGroup()`. The six nature departments are named constants (`_departmentEE` … `_departmentAL`) next to the VAT rate groups, and the dedicated `printRecItem` branch in `GenerateItems` (hardcoded 11, same output as the default branch) is gone. Generated XML is unchanged.

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
accodev: Same fix as above (24240369): plain `p.GetVatGroup()`, which knows the voucher department now.

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
accodev: Done in 24240369: the "sconto a pagare" for buoni multiuso lives in one static `EpsonPaymentType _multiUseVoucherPayment`, used both by the pay item `0x06` mapping in `GetEpsonPaymentType` and by the redeemed charge item here, so the two spellings cannot drift apart.

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
```diff
@@ -721,6 +713,23 @@ public static List<PrintRecSubtotalAdjustment> GetSubtotalAdjustments(ReceiptReq
         public static List<TotalAndMessage> GetTotalAndMessages(ReceiptRequest request)
         {
             var totalAndMessages = new List<TotalAndMessage>();
+
+            // A multi-use voucher redeemed as a charge item is paid like pay item 0x06. It goes before the other
+            // payments because a "sconto a pagare" above the amount still due is rejected with error 21.
+            foreach (var voucher in request.cbChargeItems.Where(x => x.IsMultiUseVoucherRedeem(request)))
```
mttgennarift: The zero-cash fallback at line 751 is keyed on `totalAndMessages.Count == 0`, which used to be an indirect way of asking "are there no pay items?" — the list could only be filled by `cbPayItems`. This loop adds a second writer, so the proxy no longer answers that question.

With charge items 100 + voucher −20 and **no pay items**, the list now holds one entry, the fallback does not fire, and the payment phase ends having settled 20 of 100:

```xml
<printRecItem description="Food" quantity="1" unitPrice="100" department="1" />
<printRecTotal description="Gutschein" paymentType="6" index="1" payment="20" />
<!-- 80 still due -->
```

Not a regression — that input failed before this PR too, on the negative item line — so this is a residual gap rather than something the PR broke.

Note that the fix is *not* simply keying the fallback on `cbPayItems`: that would re-add `paymentType="0" payment="0"` to the "voucher covers the whole receipt" case, which you validated on hardware (doc 6) **without** it. The condition wants to be "is there anything still due", e.g.

```csharp
var items    = (request.cbChargeItems ?? []).Where(x => !x.IsMultiUseVoucherRedeem(request)).Sum(x => Math.Abs(x.Amount));
var vouchers = (request.cbChargeItems ?? []).Where(x =>  x.IsMultiUseVoucherRedeem(request)).Sum(x => Math.Abs(x.Amount));
if (request.cbPayItems?.Any() != true && vouchers < items) { /* zero-cash fallback */ }
```

with subtotal adjustments still to be folded into `items`. Given the input is a malformed receipt anyway, documenting the limit under "Out of scope" would also be a fine answer — but the current condition reads as if it still meant "no payments given", and it doesn't.

Unrelated but adjacent: `foreach (var pay in request.cbPayItems)` below has no null guard, so a null `cbPayItems` throws today. Pre-existing, not yours — just worth covering if you touch the condition.

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
```diff
@@ -490,9 +491,11 @@ public static List<PrintRecVoid> GetRecvoids(ReceiptRequest receiptRequest)
         public static List<ItemAndMessage> GetItemAndMessages(ReceiptRequest receiptRequest)
         {
             var itemAndMessages = new List<ItemAndMessage>();
+            // A redeemed multi-use voucher is a "sconto a pagare" payment, not a line item (see GetTotalAndMessages).
+            var chargeItems = receiptRequest.cbChargeItems.Where(x => !x.IsMultiUseVoucherRedeem(receiptRequest)).ToList();
```
mttgennarift, line 495: The filter runs *before* the `GroupBy` below, so a group can now survive without its main item. `mainItem` at line 501 is a `FirstOrDefault(x => x.Position % 100 == 0)` that is dereferenced immediately (`mainItem.Quantity == 0`), so if the redeem line is the one at the multiple-of-100 position and the group has other members, the group still exists but has no head:

```
Position 200 → "Gutschein" −20   ← filtered out here
Position 201 → "Coperto"     2   ← group 2 survives, mainItem == null → NRE
```

The tests don't reach it because the voucher at 200 is alone (the whole group disappears) and the second one sits at 101 (a child, head 100 intact). Unlikely input — nobody makes a voucher the head of a grouped item — and the unguarded `FirstOrDefault` was already fragile before this PR, so an `if (mainItem is null) { continue; }` or a follow-up ticket both seem reasonable.

mttgennarift (commented):
Reviewed the shape of the fix and it looks right: mapping the redeem onto `printRecTotal paymentType="6" index="1"` (the very command pay item `0x06` already produces), emitting it before the pay items, and doing it in `GetTotalAndMessages` so invoice/refund/void are covered by one change. The `GetVatGroup` fix for the voucher *sale* line (dept 11 instead of `-1`) is a nice catch, the tests are deterministic (fixed dates, no `DateTime.Now`), and the hardware table answers the questions a reviewer would otherwise have to ask. CI is green.

Three things worth a look before merge, none of which I would call a blocker. Two are inline; the third is in a file this PR does not touch:

### Network-error recovery no longer recognises these receipts

`EpsonRTPrinterSCU.cs:369-375` (`TryRecoverFromNetworkErrorAsync`) matches an already-printed document by amount:

```csharp
var expectedAmountCents = (long) Math.Round((receiptRequest.cbReceiptAmount ?? 0) * 100);
...
var matchesByAmount = !_lastSuccessfulDoc.HasValue && docBeforeRetry != null
    && docBeforeRetry.IsFiscalDocument && docBeforeRetry.TotalDocAmountCents == expectedAmountCents;
```

For the receipt in this PR's example, `cbReceiptAmount` is the sum of the charge items, i.e. **80** (100 food − 20 voucher), while the printer's document total is **100,00** — exactly what your hardware table reports for Z 98 doc 1. So on a `TaskCanceledException` *after* the printer has printed, with `_lastSuccessfulDoc` still empty (first receipt after a restart), `matchesByAmount` is false and `RetryReceiptWithRecoveryAsync` re-sends the receipt: **duplicate document**.

Narrow window, and it needs a network fault to trigger, but the expected total should probably add back the redeemed vouchers (the same `IsMultiUseVoucherRedeem` sum used in `GetTotalAndMessages`) so that recovery keeps working for this spelling. Since the redeem now leaves the item list entirely, this is the one place where the 80-vs-100 gap becomes observable outside the printer.

### Minor

- `_departmentEE` … `_departmentAL` are `private static readonly int`; they are compile-time constants, so `const int` states the intent (and the rest of the file's `_vatRate*` fields really are mutable state, so `readonly` here reads as if these were too).
- The `GetVatGroup` short-circuit applies to *any* multi-use voucher, so it also overrides an explicitly declared nature nibble (e.g. `0x1000`/NI). The reasoning is sound (a buono multiuso is N2 whatever the POS says) — worth spelling out in the comment that the nature bits are deliberately ignored, otherwise the next reader will read it as a fall-through.
- The labels include `scu-it-epsonrtserver`, `scu-it-customrtserver` and `scu-it-customrtprinter`, but those SCUs are untouched (declared out of scope, correctly). If the follow-up tickets don't exist yet, they're worth opening now: as your description notes, there the negative `0x…48` line is printed as a *positive* NS sale line via `Math.Abs`, so the total is silently wrong instead of erroring — which is the worse failure mode of the two.

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
accodev: Good catch. Fixed in dc9a4c92: when a group has no head (the head was a redeemed voucher, or the POS never sent one) its members are printed as plain items via `GenerateItems`, so nothing that was sold is dropped. I preferred that to `continue`, which would have dropped the children silently and left the document total below what was paid. New test `Sale_GroupingRequest_VoucherAsGroupHead_PrintsTheChildrenAsItems` covers your exact shape (voucher at 200, "Coperto" at 201).

scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/Utilities/EpsonCommandFactory.cs
accodev: Agreed on both counts. Fixed in 6e15239f: the fallback is now keyed on "is anything still due" — no pay items *and* (no redeemed vouchers, or redeemed < receipt total, the signed sum of the remaining charge items). `redeemed == 0` keeps today's behaviour for every receipt without vouchers, and the "voucher covers the whole receipt" case still gets no cash total (hardware doc 6 unchanged). Validated on the sandbox printer: Food 100, voucher −20, no pay items → `paymentType="6" payment="20"` + `paymentType="0" payment="0"`, accepted as Z 98 doc 7 with total 100,00 (the printer takes the rest in cash). `cbPayItems` is null-safe on the way, as you suggested. Tests: `Sale_NoPayItems_VoucherBelowTheTotal_PaysTheRestInCash`, `Sale_NoPayItems_NoVoucher_StillPaysInCash`.

accodev:
Thanks for the thorough review, @mttgennarift. Point by point:

**Network-error recovery (80 vs 100).** The `matchesByAmount` / `expectedAmountCents` logic you quote no longer exists on `main`: #763 replaced it. `TryRecoverFromNetworkErrorAsync` now compares the printer's document counter with a baseline taken *before* the receipt is sent (`IsDocAdvanced`), and refuses to reprint when either side is missing. The amount (`TotalDocAmountCents`) is only logged, never compared, so the 80-vs-100 gap of this spelling cannot cause a duplicate. No change made.

**Group without a head** → fixed in dc9a4c92 (members printed as plain items, test added). Details in the thread.

**Zero-cash fallback** → fixed in 6e15239f (keyed on the amount still due, validated on hardware as Z 98 doc 7, tests added). Details in the thread.

**`const int` and the nature-bits comment** → 47c6feec.

**Labels.** The `scu-it-epsonrtserver` / `scu-it-customrt*` labels are added by `actions/labeler`: `.github/labeler.yml` maps the shared `fiskaltrust.Middleware.SCU.IT.Abstraction/**` folder to all four SCU labels, and the workflow also runs on `unlabeled`, so removing them just brings them back. Those SCUs are untouched here.

**Follow-up tickets** → #771 (one umbrella issue for EpsonRTServer, CustomRTPrinter and CustomRTServer, with the per-SCU findings and this PR as the reference mapping). Linked from the "Out of scope" section.

mttgennarift (approved):

PR #776, opened 2026-09-10
fix(scu-it/epson): retry the status query during recovery instead of asking once

mttgennarift:
Follow-up to #763. That change made the network-error recovery reachable; this one makes its
answer arrive in time, and stops it concluding from silence.

## Why

The recovery asked the printer what it did exactly once, right after the command wait had elapsed,
and treated "no answer" the same as "the counter did not move". Two consequences.

**The answer arrived too late to be useful.** Its cost is the full command wait plus one status
round trip, because the status query only goes out once the wait is over. Whoever calls the SCU has
a deadline of its own, and when that deadline is shorter the recovery still reaches the right
conclusion — with nobody left to hand it to. A relayed deployment measured it end to end: the
caller gave up at 100s, the recovery concluded correctly at 124s, and the document it had found was
discarded. With those two numbers the recovery could never be observed, in any run.

**And a single unanswered query ended the matter.** A printer serves one command at a time, so
while it is emitting a document it cannot reply at all. The first status query is exactly when
silence is most likely, and silence does not mean the document was not printed.

## What changes

`AwaitPrinterVerdictAsync` replaces the one-shot check with a loop that concludes only from an
answer:

| reading | conclusion | action |
|---|---|---|
| answered, counter moved | the document is on paper | recover it, no reprint |
| answered, counter unmoved | answering means the printer is not mid-document, so this is settled | nothing was emitted; sending the receipt again is safe |
| never answered in the window | unknown | `epson-printer-network-error-unknown-document-state`, never a reprint |

That middle row is what makes a short wait safe rather than dangerous: without it, shortening the
wait would only raise the chance of misreading a busy printer.

**A status query now carries its own deadline.** The loop was meant to ask several times but could
only ever ask once: `ReadLastEmittedDocStatusAsync` goes through the same client as a print command
and inherited its wait, so a silent printer — the case the loop exists for — consumed the whole
window on the first attempt. With relayed defaults the arithmetic was plain: a 45s wait per query
against a 40s window is one attempt, never two. `SendCommandAsync` takes an optional per-command
timeout and the status query passes `RecoveryStatusQueryTimeoutMs` (8s), so four attempts fit in the
window instead of none. `LocalEpsonFpMateClient` applies it as a cancellation on that one request,
surfacing as `TaskCanceledException` exactly like its own timeout, so callers cannot tell the two
apart.

New configuration, all with defaults, all per device:

- `RecoveryVerdictTimeoutMs` (40s) — how long the recovery keeps asking
- `RecoveryVerdictPollIntervalMs` (3s) — pause between questions
- `RecoveryStatusQueryTimeoutMs` (8s) — deadline of one question
- `RelayCommandTimeoutMs` (45s) — how long a command waits when the device is reached through a
  relay rather than a direct connection; ignored by `LocalEpsonFpMateClient`, which keeps using
  `ClientTimeoutMs`

## Also here

Both recovery paths signed `RTCustomerID` as empty while the happy path uses `GetCustomerTaxId`. A
receipt for a customer who gave a codice fiscale or partita IVA lost it precisely when the document
had to be reconstructed — the document on paper carries the value, because it was in the command
that printed it.

## Note for review

The optional parameter on `SendCommandAsync` breaks Moq expression trees, so five existing setups in
`DailyClosingRebootTests` and `ReprintSerialNumberTests` needed updating. Mechanical, but they are
not otherwise part of this change.

## Testing

`dotnet build scu-it/fiskaltrust.Middleware.SCU.IT.sln` — 0 errors, all three TFMs
(`netstandard2.0`, `net461`, `net6.0`).

- `EpsonRTPrinter.UnitTest` 58 passed
- `SCU.IT.UnitTest` 100 passed

The consumer-side tests live with the downstream repository that builds this SCU from source and
drives the recovery through a simulated relay. They cover the loop waiting out a printer that is
silent while it prints, the three verdicts, and — asserting the deadline rather than the behaviour,
since a fake answers instantly and hides the cost of silence — that a status query is issued with
its own short timeout.

### DIFF
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/EpsonRTPrinterSCU.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/EpsonRTPrinterSCU.cs
@@ -5,6 +5,7 @@ using fiskaltrust.Middleware.SCU.IT.Abstraction.Validation;
 using fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Models;
 using fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter.Utilities;
 using Newtonsoft.Json;
+using System.Diagnostics;
 using System.Globalization;
 using fiskaltrust.ifPOS.v1;
 using Microsoft.Extensions.Logging;
@@ -457,10 +458,20 @@ public sealed class EpsonRTPrinterSCU : LegacySCU
     /// command timeout before learning what happened, so the recovery can conclude while whoever asked for the
     /// receipt is still listening.
     /// </para>
+    /// <para>
+    /// <b>What an <see cref="IEpsonFpMateClient"/> has to guarantee for this to hold.</b> "The printer answered,
+    /// so it is not mid-document" is the assumption the resend rests on, and it is only self-evident on a direct
+    /// connection. Any client that queues commands instead of handing them straight to the device must ensure an
+    /// abandoned print command can never be delivered <em>after</em> a later status query was answered: that
+    /// ordering would let a <see cref="PrinterVerdict.NotPrinted"/> verdict be followed by the very document it
+    /// declared absent, and the resend would then duplicate it. Holding at most one command in flight per device
+    /// and discarding it when its wait elapses is sufficient, and is what the relay client does.
+    /// </para>
     /// </summary>
     private async Task<(PrinterVerdict Verdict, LastEmittedDocStatus? Doc)> AwaitPrinterVerdictAsync(string receiptReference, DocPosition baseline)
     {
-        var deadline = DateTime.UtcNow.AddMilliseconds(_configuration.RecoveryVerdictTimeoutMs);
+        // Stopwatch rather than the wall clock: a clock jump must not shorten or extend the window.
+        var elapsed = Stopwatch.StartNew();
         var attempt = 0;
 
         while (true)
@@ -481,7 +492,7 @@ public sealed class EpsonRTPrinterSCU : LegacySCU
                 return (PrinterVerdict.NotPrinted, doc);
             }
 
-            if (DateTime.UtcNow >= deadline)
+            if (elapsed.ElapsedMilliseconds >= _configuration.RecoveryVerdictTimeoutMs)
             {
                 _logger.LogError("({receiptreference}) Printer did not answer within {timeout}ms ({attempts} attempts) — state unknown.",
                     receiptReference, _configuration.RecoveryVerdictTimeoutMs, attempt);
--- a/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/EpsonRTPrinterSCU.cs
+++ b/scu-it/src/fiskaltrust.Middleware.SCU.IT.EpsonRTPrinter/EpsonRTPrinterSCU.cs
@@ -431,34 +431,96 @@ public sealed class EpsonRTPrinterSCU : LegacySCU
         _lastDoc.Value = null;
     }
 
+    private enum PrinterVerdict
+    {
+        /// <summary>The printer answered and its counter had moved: the document is on paper.</summary>
+        Printed,
+
+        /// <summary>The printer answered and its counter had not moved: nothing was emitted.</summary>
+        NotPrinted,
+
+        /// <summary>The printer never answered within the window. Nothing can be concluded.</summary>
+        Unknown
+    }
+
+    /// <summary>
+    /// Asks the printer what it did, repeatedly, until it answers or the window closes.
+    /// <para>
+    /// The single rule that makes this safe: a verdict is drawn only from an answer, never from silence. A
+    /// printer serves one command at a time, so it cannot reply to a status query while it is emitting a
+    /// document — an answer therefore means the operation is over and its counter is a settled fact. Silence,
+    /// on the other hand, is indistinguishable between "still printing" and "gone", which is why it ends in
+    /// <see cref="PrinterVerdict.Unknown"/> and never in a reprint.
+    /// </para>
+    /// <para>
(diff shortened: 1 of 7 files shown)


danielscainifiskaltrust:
## Review

**Verdict: approvable, with one important reservation on the consumer (relay) side and a suggestion on tests.** The logic is correct and well documented, and the `RTCustomerID` fix is a real improvement. No blocking bugs in this repo's code.

### Verified points (all good)

- **The loop cannot throw from the query**: `ReadLastEmittedDocStatusAsync` catches everything and returns `null`, so the `TaskCanceledException` from the short per-query timeout correctly becomes "no answer, ask again". At least one attempt is guaranteed because the deadline check runs after the query.
- **The stated arithmetic checks out**: with an 8s query + 3s pause inside a 40s window you get 4 attempts, as the description says.
- **The `RTCustomerID` fix is consistent with the happy path**: the happy path uses exactly `GetLotteryCode`/`GetCustomerTaxId`, and changing `ApplyRecoveredDoc` to take the `ReceiptRequest` also removes the lottery-code-only parameter threading. The `baseline == null` semantics are preserved (immediate error, as before).
- **`LocalEpsonFpMateClient`**: the per-request CTS is correct; `HttpClient.Timeout` still applies as the upper bound and cancellation surfaces as `TaskCanceledException` on all TFMs, indistinguishable from the client's own timeout as claimed.
- The Moq updates in the existing tests are indeed purely mechanical.

### Main reservation — the relay side is not in this PR, but it is the case that motivates it

1. **`RelayCommandTimeoutMs` is not read anywhere in this repo** — it is only consumed by the downstream relay client. Downstream, `EpsonCloudHttpClient` still has the old `SendCommandAsync(string)` signature and a hardcoded 120s command timeout. When the submodule is bumped: (a) the downstream build **breaks**, because the implementation no longer satisfies the interface with the `TimeSpan?` parameter; (b) more importantly, if the relay client does not **honor** the per-command timeout, the status query in a relayed deployment — exactly the scenario that motivated this PR — will still burn the whole recovery window on the first attempt. The companion PR in the downstream repo should be ready before the submodule bump; please link it in the description.
2. **The "an answer means the counter is settled" invariant across the relay**: locally it holds (the printer serves one command at a time). Through the relay it only holds if delivery is FIFO on the same channel — if the print command could be delivered *after* the status query has been answered, a `NotPrinted` verdict would produce a fiscal duplicate (we have already seen a late relay delivery cause a duplicate, in August). With the relay's FIFO queue the invariant holds; it is worth stating explicitly in a comment or in the relay client's docs, because it is the load-bearing assumption behind the resend.

### Minor observations (non-blocking)

- **No unit tests in this repo for the new logic**: `AwaitPrinterVerdictAsync` lives here, but the tests for the three verdicts only exist downstream. With a mocked `IEpsonFpMateClient` (silent-then-answering) and poll intervals set to 0–1ms via config, the loop is testable here too; I would suggest adding that, since the middleware repo is the canonical upstream.
- **The window can overrun**: the last query may start just before the deadline and cost another 8s (worst case ~40+3+8s). Harmless with the defaults, but the docs say a flat "40s".
- **Overall retry budget**: the 100s caller math covers the first pass (45+40 = 85s), but with `NotPrinted` → resend → another network error, each attempt adds up to 45+40s; with `MaxNetworkRetries = 3` the worst case far exceeds the caller's deadline. Not a regression (it was worse before), but the "the verdict arrives while the caller is still listening" guarantee only holds for the first pass.
- `DateTime.UtcNow` for the deadline: a `Stopwatch` would be immune to clock jumps. Nit.

In short: the PR's content is solid and merging is safe for the middleware repo itself; I would however explicitly coordinate it with the downstream companion change, because without it the fix is neither effective (nor compilable) in precisely the relayed deployment that originated it.

mttgennarift:
Thanks — the reservation about the consumer side is the right one to raise, and the invariant question is the sharpest point in the review. Addressed in 7aa8b7a1.

**Tests here.** Agreed that upstream is the canonical place for logic that lives upstream. `RecoveryVerdictTests` now covers the three verdicts, a printer that is silent while it prints and answers later, a non-fiscal last document, the recovered customer tax id, and the per-query deadline — that last one asserted on the deadline rather than on behaviour, precisely because a mock answers instantly and cannot show what silence costs.

**The invariant.** Your framing is right, and the guarantee is slightly stronger than FIFO. The relay keeps at most one command in flight per device, and when a command's wait elapses it is *removed* before the failure is raised — so an abandoned print command is not queued behind the status query, it is gone, and cannot be delivered late. The two remaining cases are: the device never fetched it (it is discarded, nothing was printed), or the device fetched it and is printing (it will not poll for the status query until it is done, by which point the counter reflects the outcome). `AwaitPrinterVerdictAsync` now states this as a requirement on any `IEpsonFpMateClient` implementation rather than leaving it implicit, since it is what the resend rests on.

On the August duplicate: as documented it came from the POS resubmitting the sale days later with a fresh operation id after a timeout, not from a late delivery over the relay. Worth separating the two, because they call for different defences — the first is what the `unknown-document-state` marker exists for.

**Window overrun and retry budget.** Both correct, both now documented on `RecoveryVerdictTimeoutMs`: it bounds when the last query may *start*, so a pass can overrun by one status query; and it is the budget of a single pass, with `MaxNetworkRetries` multiplying the worst case. We chose to document rather than cap, because `MaxNetworkRetries` is already per-device configuration — a deployment with a tight caller deadline sets it to 1 and gets one predictable pass, while one that has raised its own timeout can afford the extra attempts. Capping in code would remove that choice.

**`Stopwatch`.** Done.

**Consumer side.** The companion change is ready and does honour the per-command timeout — the relay client applies it as the deadline of that one command, which is what makes the short status query actually short in a relayed deployment. It is not linked here because it lives in a private repository; happy to share the link directly with anyone reviewing. The order is as you say: it must merge, and the submodule pointer bump must follow, before a relayed deployment sees either the fix or a compiling build.

accodev (approved):

